PR #63, opened 2015-01-04
Fix OS X problems

brucehoult:
Fix problems reported on mailing list by John Leung john at johnleung.com 

https://lists.opendylan.org/pipermail/bdwgc/2014-December/006066.html

Explanation of problems and fixes:

There are actually two problems here.

The first is in:

7bef74b Fix unresolved vsnprintf in misc.c and snprintf in cordtest (DJGPP,
VC)

OS X for some reason has problems with defining snprintf as a macro and
including another macro expansion in its arguments. Others have found this
in the past e.g.

http://sourceforge.net/p/powerwatershed/discussion/1024808/thread/79db23d3/

The next problem is in:

0d147af Fix missing error handling of pthread_attr_init/getstacksize

Error 22 is EINVAL.  Man pthread_attr_setstacksize gives the following
possible reasons for EINVAL:

     [EINVAL]           Invalid value for attr.
     [EINVAL]           stacksize is less than PTHREAD_STACK_MIN.
     [EINVAL]           stacksize is not a multiple of the system page size.

Changing the requested stack size from 1000000 to 1000*1024 fixes the problem, demonstrating that the issue is "stacksize is not a multiple of the system page size."

### DIFF
--- a/cord/tests/cordtest.c
+++ b/cord/tests/cordtest.c
@@ -206,9 +206,9 @@ void test_extras(void)
 
 #if defined(__DJGPP__) || defined(__STRICT_ANSI__)
   /* snprintf is missing in DJGPP (v2.0.3) */
-# define GC_SNPRINTF sprintf
-# define GC_SNPRINTF_BUFSZ_ARG(bufsz) /* empty */
+# define HAVE_SNPRINTF 0
 #else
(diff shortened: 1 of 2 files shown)


ivmai:
Thank you. I merged it (slightly modified) to master

PR #68, opened 2015-06-25
Fix build for some iOS targets

biochimia:
Fix compilation on recent iOS versions, 32-bit mode and enable compilation of the collector when targeting the iPhone simulator.


ivmai:
Thank you
The fixes cherry-picked to master (commit messages adjusted), see cb373bb and 8d4fc8d

biochimia:
Thanks for taking this in and for fixing the commit messages.

PR #80, opened 2015-12-04
Thread local allocation for typed objects

hoheinzollern:
Hi Ivan,
I am working for Prolog Development Center in an effort to improve the performance of multi-threaded code. We found that the thread local storage was organized a bit differently from the global storage, and hence we refactored some of it so that the global and the local storages would be similar to eachother, and moved some repeated functions into gc_inline.h. Finally, we have implemented the thread local version of the typed allocation.

We successfully tested the code and it is working well in our configuration, compiled under Visual Studio 2013. We also tried to ensure that the code would compile under all other combinations of flags that we touched, but we have not tried to run it in those cases.

We would be very happy if you could review the changes and possibly accept them, and have in our queue two other requests: a stress test for the garbage collector, which I first need to convert to C (you can find the C++ version in my repository) and possibly a memory profiler.

### DIFF
--- a/include/gc_inline.h
+++ b/include/gc_inline.h
@@ -47,6 +47,16 @@
 /* traced even if objects are ptrfree.                                  */
 GC_API void GC_CALL GC_generic_malloc_many(size_t /* lb */, int /* k */,
                                            void ** /* result */);
+/* Generic allocation function.  Chooses appropriately the thread local */
+/* or the global freelist for the correct kind, then returns a pointer  */
+/* to the alloc'd memory                                                */
+GC_API void * GC_CALL GC_generic_malloc_kind(size_t /* lb */,
+                                             int /* k */);
+#ifdef THREAD_LOCAL_ALLOC
+/* Define the global version in case thread local allocation is enabled */
+GC_API void * GC_CALL GC_generic_malloc_kind_global(size_t /* lb */,
+                                                    int /* k */);
+#endif
 
 /* The ultimately general inline allocation macro.  Allocate an object  */
 /* of size granules, putting the resulting pointer in result.  Tiny_fl  */
--- a/mallocx.c
+++ b/mallocx.c
@@ -39,12 +39,12 @@
 /* Some externally visible but unadvertised variables to allow access to */
 /* free lists from inlined allocators without including gc_priv.h        */
 /* or introducing dependencies on internal data structure layouts.       */
-void ** const GC_objfreelist_ptr = GC_objfreelist;
-void ** const GC_aobjfreelist_ptr = GC_aobjfreelist;
-void ** const GC_uobjfreelist_ptr = GC_uobjfreelist;
-# ifdef GC_ATOMIC_UNCOLLECTABLE
-    void ** const GC_auobjfreelist_ptr = GC_auobjfreelist;
-# endif
+    void ** const GC_objfreelist_ptr = GC_freelist[NORMAL];
+    void ** const GC_aobjfreelist_ptr = GC_freelist[PTRFREE];
+    void ** const GC_uobjfreelist_ptr = GC_freelist[UNCOLLECTABLE];
+#   ifdef GC_ATOMIC_UNCOLLECTABLE
+      void ** const GC_auobjfreelist_ptr = GC_freelist[AUNCOLLECTABLE];
+#   endif
 
 GC_API int GC_CALL GC_get_kind_and_size(const void * p, size_t * psize)
 {
--- a/thread_local_alloc.c
+++ b/thread_local_alloc.c
@@ -162,19 +162,19 @@ GC_INNER void GC_destroy_thread_local(GC_tlfs p)
   GC_bool GC_is_thread_tsd_valid(void *tsd);
 #endif
 
-GC_API GC_ATTR_MALLOC void * GC_CALL GC_malloc(size_t bytes)
+#ifdef THREAD_LOCAL_ALLOC
+  GC_API void * GC_generic_malloc_kind(size_t bytes, int kind)
 {
     size_t granules = ROUNDED_UP_GRANULES(bytes);
     void *tsd;
     void *result;
     void **tiny_fl;
-
 #   if !defined(USE_PTHREAD_SPECIFIC) && !defined(USE_WIN32_SPECIFIC)
       GC_key_t k = GC_thread_key;
       if (EXPECT(0 == k, FALSE)) {
         /* We haven't yet run GC_init_parallel.  That means     */
         /* we also aren't locking, so this is fairly cheap.     */
-        return GC_core_malloc(bytes);
+        return GC_generic_malloc_kind_global(bytes, kind);
       }
       tsd = GC_getspecific(k);
 #   else
@@ -182,58 +182,21 @@ GC_API GC_ATTR_MALLOC void * GC_CALL GC_malloc(size_t bytes)
 #   endif
 #   if !defined(USE_COMPILER_TLS) && !defined(USE_WIN32_COMPILER_TLS)
       if (EXPECT(0 == tsd, FALSE)) {
-        return GC_core_malloc(bytes);
+        return GC_generic_malloc_kind_global(bytes, kind);
       }
 #   endif
     GC_ASSERT(GC_is_initialized);
-
     GC_ASSERT(GC_is_thread_tsd_valid(tsd));
-
-    tiny_fl = ((GC_tlfs)tsd) -> freelists[NORMAL];
+    tiny_fl = ((GC_tlfs)tsd)->freelists[kind];
     GC_FAST_MALLOC_GRANS(result, granules, tiny_fl, DIRECT_GRANULES,
-                         NORMAL, GC_core_malloc(bytes), obj_link(result)=0);
-#   ifdef LOG_ALLOCS
-      GC_log_printf("GC_malloc(%lu) returned %p, recent GC #%lu\n",
-                    (unsigned long)bytes, result, (unsigned long)GC_gc_no);
-#   endif
+        kind, GC_generic_malloc_kind_global(bytes, kind), obj_link(result) = 0);
     return result;
 }
-
-GC_API GC_ATTR_MALLOC void * GC_CALL GC_malloc_atomic(size_t bytes)
-{
-    size_t granules = ROUNDED_UP_GRANULES(bytes);
-    void *tsd;
-    void *result;
-    void **tiny_fl;
-
-#   if !defined(USE_PTHREAD_SPECIFIC) && !defined(USE_WIN32_SPECIFIC)
-      GC_key_t k = GC_thread_key;
-      if (EXPECT(0 == k, FALSE)) {
-        /* We haven't yet run GC_init_parallel.  That means     */
-        /* we also aren't locking, so this is fairly cheap.     */
-        return GC_core_malloc_atomic(bytes);
-      }
-      tsd = GC_getspecific(k);
-#   else
-      tsd = GC_getspecific(GC_thread_key);
-#   endif
-#   if !defined(USE_COMPILER_TLS) && !defined(USE_WIN32_COMPILER_TLS)
-      if (EXPECT(0 == tsd, FALSE)) {
-        return GC_core_malloc_atomic(bytes);
-      }
-#   endif
-    GC_ASSERT(GC_is_initialized);
-    tiny_fl = ((GC_tlfs)tsd) -> freelists[PTRFREE];
-    GC_FAST_MALLOC_GRANS(result, granules, tiny_fl, DIRECT_GRANULES, PTRFREE,
-                         GC_core_malloc_atomic(bytes), (void)0 /* no init */);
-    return result;
-}
+#endif /* THREAD_LOCAL_ALLOC */
 
 #ifdef GC_GCJ_SUPPORT
 
 # include "atomic_ops.h" /* for AO_compiler_barrier() */
-
-# include "include/gc_gcj.h"
 
 /* Gcj-style allocation without locks is extremely tricky.  The         */
 /* fundamental issue is that we may end up marking a free list, which   */
--- a/thread_local_alloc.c
+++ b/thread_local_alloc.c
@@ -12,6 +12,11 @@
  */
 
 #include "private/gc_priv.h"
+#include "gc_inline.h"
+
+#ifdef GC_GCJ_SUPPORT
+# include "gc_gcj.h"
+#endif
 
 #if defined(THREAD_LOCAL_ALLOC)
 
@@ -99,7 +104,7 @@ static void return_freelists(void **fl, void **gfl)
 /* This call must be made from the new thread.  */
 GC_INNER void GC_init_thread_local(GC_tlfs p)
 {
-    int i;
+    int i, j;
 
     GC_ASSERT(I_HOLD_LOCK());
     if (!EXPECT(keys_initialized, TRUE)) {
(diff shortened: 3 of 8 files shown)


ivmai:
Thank you for proposing an improvement. I will definitely review it and accept provided it does not break anything. There could be multiple rounds of code review (involving other people as well). The whole process could take a month or longer (so please be prepared for).
Parallel pull requests are OK.
It is of course better to break big commits into smaller ones (provided they could be logically divided into).
Since it is not a trivial patch please duplicate your request (with the description) to the GC mailing list (bdwgc@lists.opendylan.org).

include/gc_inline.h
```diff
@@ -28,6 +28,8 @@
 /* It is also used internally for thread-local allocation.              */
 /* Manual use is hereby discouraged.                                    */
 
+#include "private/gc_priv.h"
```
ivmai: Don't include private headers from public ones.

ivmai:
1. Commit message is too short - please describe it in details
2. The commit is quite big - is it possible to break it into smaller (logically separated) ones?

win32_threads.c
```diff
@@ -344,7 +344,7 @@ STATIC GC_thread GC_new_thread(DWORD id)
   } else {
     GC_ASSERT(!GC_win32_dll_threads);
     result = (struct GC_Thread_Rep *)
-                GC_INTERNAL_MALLOC(sizeof(struct GC_Thread_Rep), NORMAL);
```
ivmai: delete this change

win32_threads.c
```diff
@@ -2866,3 +2866,4 @@ GC_INNER void GC_init_parallel(void)
 # endif /* !GC_NO_THREAD_REDIRECTS */
 
 #endif /* GC_WIN32_THREADS */
```
ivmai: delete this change

typd_mlc.c
```diff
@@ -665,8 +643,7 @@ GC_API GC_ATTR_MALLOC void * GC_CALL
    return((void *) op);
 }
```
ivmai: delete this change

include/gc_inline.h
```diff
@@ -143,4 +145,72 @@ GC_API void GC_CALL GC_generic_malloc_many(size_t /* lb */, int /* k */,
     ((void **)(result))[1] = (void *)(second); \
   } while (0)
```
ivmai: THREAD_LOCAL_ALLOC and GC_INLINE cannot be in public header

mallocx.c
```diff
@@ -39,11 +39,11 @@
 /* Some externally visible but unadvertised variables to allow access to */
 /* free lists from inlined allocators without including gc_priv.h        */
 /* or introducing dependencies on internal data structure layouts.       */
-void ** const GC_objfreelist_ptr = GC_objfreelist;
-void ** const GC_aobjfreelist_ptr = GC_aobjfreelist;
-void ** const GC_uobjfreelist_ptr = GC_uobjfreelist;
+void ** const GC_objfreelist_ptr = GC_freelist[NORMAL];
+void ** const GC_aobjfreelist_ptr = GC_freelist[PTRFREE];
+void ** const GC_uobjfreelist_ptr = GC_freelist[UNCOLLECTABLE];
 # ifdef GC_ATOMIC_UNCOLLECTABLE
```
ivmai: wrong indentation, void should start at position 4

thread_local_alloc.c
```diff
@@ -75,7 +81,7 @@ static void return_freelists(void **fl, void **gfl)
     }
     /* The 0 granule freelist really contains 1 granule objects.        */
 #   ifdef GC_GCJ_SUPPORT
```
ivmai: wrong indentation of "if"

thread_local_alloc.c
```diff
@@ -99,7 +105,7 @@ static void return_freelists(void **fl, void **gfl)
 /* This call must be made from the new thread.  */
 GC_INNER void GC_init_thread_local(GC_tlfs p)
 {
```
ivmai: wrong indent of int

thread_local_alloc.c
```diff
@@ -112,124 +118,51 @@ GC_INNER void GC_init_thread_local(GC_tlfs p)
...
+        }
+        /* Set up the size 0 free lists.    */
+        /* We now handle most of them like regular free lists, to ensure    */
+        /* That explicit deallocation works.  However, allocation of a      */
+        /* size 0 "gcj" object is always an error.                          */
+#   ifdef GC_GCJ_SUPPORT
+        if (i == GC_gcj_kind) {
+          p->freelists[i][0] = ERROR_FL;
+        } else {
+#   endif
```
ivmai: change code to (with the proper indentation):
} else
\#endif
/* else */ {
p -> ...
}

thread_local_alloc.c
```diff
@@ -112,124 +118,51 @@ GC_INNER void GC_init_thread_local(GC_tlfs p)
...
+    for (i = 0; i < TINY_FREELISTS; ++i) {
+        p -> finalized_freelists[i] = (void *)(word)1;
+    }
+    p->finalized_freelists[0] = (void *)(word)1;
 #   endif
 }
 
 /* We hold the allocator lock.  */
 GC_INNER void GC_destroy_thread_local(GC_tlfs p)
 {
```
ivmai: restore original indentation

thread_local_alloc.c
```diff
@@ -251,7 +184,7 @@ GC_API GC_ATTR_MALLOC void * GC_CALL GC_malloc_atomic(size_t bytes)
 /* Unlike the other thread local allocation calls, we assume that the   */
 /* collector has been explicitly initialized.                           */
 GC_API GC_ATTR_MALLOC void * GC_CALL GC_gcj_malloc(size_t bytes,
-                                    void * ptr_to_struct_containing_descr)
```
ivmai: revert this change

thread_local_alloc.c
```diff
@@ -259,14 +192,14 @@ GC_API GC_ATTR_MALLOC void * GC_CALL GC_gcj_malloc(size_t bytes,
...
     void *result;
     void **tiny_fl = ((GC_tlfs)GC_getspecific(GC_thread_key))
-                                        -> gcj_freelists;
+      -> freelists[GC_gcj_kind];
     GC_ASSERT(GC_gcj_malloc_initialized);
     GC_FAST_MALLOC_GRANS(result, granules, tiny_fl, DIRECT_GRANULES,
                          GC_gcj_kind,
                          GC_core_gcj_malloc(bytes,
                                             ptr_to_struct_containing_descr),
                          {AO_compiler_barrier();
```
ivmai: revert the change

thread_local_alloc.c
```diff
@@ -298,44 +231,39 @@ GC_API GC_ATTR_MALLOC void * GC_CALL GC_gcj_malloc(size_t bytes,
...
+        for (j = 0; j < TINY_FREELISTS; ++j) {
+            q = p->freelists[i][j];
 #     ifdef GC_GCJ_SUPPORT
-        if (j > 0) {
-          q = p -> gcj_freelists[j];
-          if ((word)q > HBLKSIZE) GC_set_fl_marks(q);
-        }
+            if (j > 0) {
+                if ((word)q > HBLKSIZE) GC_set_fl_marks(q);
+            } else {
```
ivmai: See code style for this above

thread_local_alloc.c
```diff
@@ -298,44 +231,39 @@ GC_API GC_ATTR_MALLOC void * GC_CALL GC_gcj_malloc(size_t bytes,
...
-#     endif
+                if ((word)q > HBLKSIZE) GC_set_fl_marks(q);
+#     ifdef GC_GCJ_SUPPORT
+            }
+#     endif /* GC_GCJ_SUPPORT */
+        }
     }
 }
 
 #if defined(GC_ASSERTIONS)
```
ivmai: Restore original indentation

typd_mlc.c
```diff
@@ -198,7 +197,7 @@ GC_make_sequence_descriptor(complex_descriptor *first,
 /* each of which can be described by a simple descriptor.       */
 /* We try to optimize some common cases.                        */
 /* If the result is COMPLEX, then a complex_descr* is returned  */
```
ivmai: revert this change

typd_mlc.c
```diff
@@ -593,40 +592,19 @@ GC_API GC_descr GC_CALL GC_make_descriptor(const GC_word * bm, size_t len)
...
-        ((word *)op)[GRANULES_TO_WORDS(lg) - 1] = d;
-   } else {
-       op = (ptr_t)GENERAL_MALLOC((word)lb, GC_explicit_kind);
-       if (op != NULL) {
-            lg = BYTES_TO_GRANULES(GC_size(op));
-            ((word *)op)[GRANULES_TO_WORDS(lg) - 1] = d;
-       }
-   }
-   return((void *) op);
+GC_API GC_ATTR_MALLOC void * GC_CALL GC_malloc_explicitly_typed(size_t bytes, GC_descr d)
```
ivmai: Restore original formatting (indentation)

typd_mlc.c
```diff
@@ -593,40 +592,19 @@ GC_API GC_descr GC_CALL GC_make_descriptor(const GC_word * bm, size_t len)
...
+{
+    bytes += TYPD_EXTRA_BYTES;
```
ivmai: declaration should precede any statement in a block (this is C89 not C++)

include/gc_inline.h
```diff
@@ -143,4 +145,72 @@ GC_API void GC_CALL GC_generic_malloc_many(size_t /* lb */, int /* k */,
...
+    if (EXPECT(0 == k, FALSE)) {
+        /* We haven't yet run GC_init_parallel.  That means     */
+        /* we also aren't locking, so this is fairly cheap.     */
+        return GC_generic_malloc_kind_global(bytes, kind);
+    }
+    tsd = GC_getspecific(k);
+#else
+    tsd = GC_getspecific(GC_thread_key);
+#endif
+#if !defined(USE_COMPILER_TLS) && !defined(USE_WIN32_COMPILER_TLS)
```
ivmai: wrong indentation of \#if - "{" increases indentation level, so here it should be \#<space>if (same else, endif below and above)

ivmai:
I have quickly glanced the commit in github web interface. All my comments above are mostly regarding syntax (I hope you will fix it not waiting for my further comments if possible). Code review at semantic level needs more time. Actually the proposed change is quite big, at this moment I'm sure we need the whole one. Based on your description the change actually contains 3 items - please try to put them in different commits. Also, does it really make sense to make the functions inline - have you notice performance impact?

hoheinzollern:
Hi Ivan, thank you for your comments. I will work on them on Monday but I can give you a quick reply.

Yes, I can split the commit into 3:
1. changes to the thread local data structure
2. unification of some common pieces of code put in gc_inline.h
3. thread-local allocation for the typed kind.

Regarding inlining, I have not noticed a performance impact, but it helped me a lot with the clarity of the code that I was writing, as I could identify some repeated chunks that the other versions of malloc were using. The other option is to write a macro for the same purpose, but I would like to avoid that if possible. In both cases I would like to keep this general version of malloc, and maybe convert even more of the other specific versions to it, to avoid copy/paste programming mistakes.

As a final note, you said here in the comments "THREAD_LOCAL_ALLOC and GC_INLINE cannot be in public header". Could you explain this to me? Am I not supposed to use these macro definitions in gc_inline.h?
The rest of the comments I understand and agree with. Thank you again for this quick review :)

ivmai:
> As a final note, you said here in the comments "THREAD_LOCAL_ALLOC and GC_INLINE cannot be in public header". Could you explain this to me?
THREAD_LOCAL_ALLOC influences GC configuration but not the API (see doc/README.macros).
Of course, it is possible to add new client-side macro e.g. GC_INLINE_TYPED_LOCAL_ALLOC if really needed.
GC_INLINE is a macro defined and used internally (not supposed to be used in any public GC header) - it is a kind of suggestion to inline a function, e.g. it is defined to "static" if strict ANSI C compilation requested. (Currently I have no idea how to declare inline functions in public headers.)

> Regarding inlining, I have not noticed a performance impact, but it helped me a lot with the clarity of the code that I was writing, as I could identify some repeated chunks that the other versions of malloc were using. The other option is to write a macro for the same purpose
Why not to define ordinary (non-inline) function instead?

hoheinzollern:
I now split the commit into two changes instead of three, as I think it makes more sense. Basically commit #74549b0 modifies the thread local data structures, while commit #af4668a inserts the GC_malloc_generic_kind function in the thread local and global flavours, and remaps GC_malloc, GC_malloc_atomic and GC_malloc_explicit_kind to make use of it. As you suggested I defined the function as global, and removed the inclusion of private headers from the public interface.
I hope this passes your first review and the new message descriptions are detailed enough. If so we should announce this to the public mailing list.

ivmai:
Thank you, I will review them within 2 weeks.

BTW. Do you know about this - https://github.com/SWI-Prolog/misc-bdwgc ?

hoheinzollern:
Thank you. No, we didn't know that SWI-prolog was also using this Boehm's GC. It would be good to know if they are just using a vanilla copy of this repository or made any customization to it. The former seems to be true.
Although I'm not familiar with SWI-prolog internals, Visual Prolog makes use of static type annotations at compile time, in contrast to standard Prolog environments which are dynamic in nature. This could mean that we have a different usage pattern of the garbage collector, but I'm just speculating.

ivmai:
I have reviewed your 1st commit, refactored the code (mainly removed regression on DRAM usage, i.e. default array size occupies same amount of memory), adjusted commit message and cut it to 2 commits - see 41871b9 and a03bd3a
Please rebase your 2nd commit - I will review it then

include/gc_inline.h
```diff
@@ -47,6 +47,16 @@
 /* traced even if objects are ptrfree.                                  */
 GC_API void GC_CALL GC_generic_malloc_many(size_t /* lb */, int /* k */,
                                            void ** /* result */);
+/* Generic allocation function.  Chooses appropriately the thread local */
```
ivmai: Can we move these declarations to gc_mark.h?

tholp:
The problem I wrote about in 41871b970c8fd9704835c1a221a8f9f1deea707a has large influence on this.  Rebasing does not really make sense until that problem has been resolved.

hoheinzollern:
Hi Ivan, let me explain the current situation: we tried to rebase our second commit to take your changes that fix the regression on memory usage. However the problem is now that the thread_local data structure does not have enough space for a new kind, hence the test fails when we try to dynamically allocate a typed kind.

There is also an inconsistency between the structure obj_kind in gc_priv.h, the size of the _freelists matrix in gc_priv.h, and the thread_local_freelists in thread_local_alloc.h. The first is of size MAX_OBJ_KIND, the second is of size PREDEFINED_KINDS (=STUBBORN[+1]) and the last only contains THREAD_FREELISTS_KINDS (=NORMAL+1) elements.
This makes it hard to find a general rule for managing the freelists, and to decide when there is space for dynamically allocated ones.

I understand your problem with big tables of thread-local freelists, as they occupy too much memory, especially on legacy systems, which you support. But for newer ones, in particular those with on-demand allocation, reserving a good chunk of address space and then lazily initializing only the required parts would not be an issue, as is our case.
On the other hand, the dynamic creation of a new kind should fail if there is not enough space in the freelist matrix. This problem should be fixed before we move on to rebasing the second commit.

Since you are the owner of the project, we would be happy if you made a decision that you are satisfied with, and that can accomodate enough space for dynamically allocated freelists, possibly with some configuration flags.

ivmai:
I see. I have refactored your second commit (except for typed allocation) - please see 9890568
The client can specify capacity of the free lists even from the command line

tholp:
Hi, Ivan.  I still don't see the reason for so many individual "MAX" constants, but if we just define #88

ivmai:
If we decrease some of MAX values then it could break some clients (who rely on default MAX values).
If we increase some of MAX values then memory consumption (or waste) will grow, though I have not checked how much is the growth. Probably later we will increase the values and make them all the same.
But starting from now the clients could select the MAX values - either like you pointed out or specifying them in the command line (e.g. CFLAG="-D PREDEFINED_KINDS=6 -D MAXOBJKINDS=6 -D THREAD_FREELISTS_KINDS=6").

ivmai:
Typed allocation patch is in #88

PR #82, opened 2015-12-21
Silence CMake warning about CMP0054 by unquoting instances of "HOST"

andrewbuss:
CMake 3.1 changed how variables in quoted strings are substituted: 
https://cmake.org/cmake/help/v3.4/policy/CMP0054.html


ivmai:
Cherry-picked to master (see 4e78948). Thank you

Issue #70, opened 2015-06-28
Trouble with gc-7.2d compiling against musl libc


ivmai:
On behalf of John Spencer maillist-boehm@barfooze.de:

Hi Ivan,

did the results of "make check" give you any clue ?

On 01/08/2013 11:31 AM, Ivan Maidanski wrote:

> > > > > which algorithm is used for detection of main stack base in your case? which macro is defined - STACKBOTTOM, HEURISTIC1, LINUX_STACKBOTTOM?
> > > > 
> > > > out of these 3, only LINUX_STACKBOTTOM seems to be defined (i added
> > > > # warnings into os_dep.c in order to find out)
> > > > 
> > > > Seems to be correct.
> > > > Does GC_get_main_stack_base return reasonable value?
> > 
> > how can i check ?
> > What is the difference between this value and address of any local var of main() function? (It should not be bigger than several K.) Is this value bigger than that local var address? (stack grows down on your target, right?)

~/GC/bdwgc # cat test.c
# include "include/private/gc_priv.h"
# include <stdio.h>

int main() {
         int a;
         dprintf(2, "a: %p, stack: %p\n", &a, GC_get_main_stack_base());
}

~/GC/bdwgc # gcc test.c -L".libs/" -lgc
~/GC/bdwgc # LD_LIBRARY_PATH=.libs ./a.out
a: 0x7fff6ff6728c, stack: 0x7fff6ff672e0
~/GC/bdwgc #

> > > > > > in any case, I'd welcome advice on how to workaround this issue, so that
> > > > > > i get a working libgc.so in order to use inkscape.
> > > > > > Thanks for reporting the problem, I'll definitely fix the gc source if we understand how to do it correctly. (Typically, first I apply the fixes to master branch and, then, cherry-pick to 7.2 release.)
> > > > 
> > > > thank you. is there anything else i can help with in order to get this
> > > > fixed ?
> > > > Also, static data roots should be checked (GC_static_roots[], all global/static variables should be within one of the regions).
> > 
> > how can i check ? List all root ranges.
> > Compile with linker map output, it will show addresses of global variables. These addresses (excluding read-only data) must be inside of a root range.

sorry, i don't know how to do that. please provide a test program and/or
required command line.

i tried:
~/GC/bdwgc # gcc -shared -fPIC -DPIC .libs/allchblk.o .libs/alloc.o
.libs/blacklst.o .libs/checksums.o .libs/db
[...]

PR #84, opened 2015-12-30
Fix building on musl libc

somasis:
Without this fix, I needed to set -DNO_GETCONTEXT to finish the final linking.

### DIFF
--- a/include/private/gcconfig.h
+++ b/include/private/gcconfig.h
@@ -2801,7 +2801,8 @@
                              || defined(OPENBSD) || defined(ARM32) \
                              || defined(MIPS) || defined(AVR32) \
                              || defined(OR1K))) \
-     || (defined(LINUX) && (defined(SPARC) || defined(M68K))) \
+     || (defined(LINUX) && ((!defined(__GLIBC__) && !defined(__UCLIBC__)) \
+                             || defined(SPARC) || defined(M68K))) \
      || (defined(RTEMS) && defined(I386)) || defined(PLATFORM_ANDROID)) \
     && !defined(NO_GETCONTEXT)
 # define NO_GETCONTEXT


ivmai:
Fixed by other patch, see d7c2d16

somasis:
Thanks :+1:

ivmai:
So, now master can be built with musl successfully, right?
Do the tests pass? Someone complained (using an older bdwgc) that stack bottom is not properly detected, see #70 - it would be good if you check it.

somasis:
Builds fine. Five tests fail; cordtest, gctest, threadkey_test, test_cpp, and disclaim_bench. http://ix.io/nkP

ivmai:
So, bdwgc with musl is not working. Could you please try this:
gcc -O0 -g -Wall -Wextra -I include -DGC_ASSERTIONS -DGC_PRINT_VERBOSE_STATS -DGC_DISABLE_INCREMENTAL tests/test.c extra/gc.c && ./a.out

somasis:
Here's the output of a.out. http://ix.io/nmn

ivmai:
Single-threaded mode is OK. Does it really use musl libc?

somasis:
What do you mean exactly? This system is entirely musl libc. I've been
able to use programs that depend on it (inkscape and elinks are what
I've tested) and they work just fine, so I know it at least works in
application.

Multi-threaded results: http://ix.io/nqV

ivmai:
Results file is incomplete (no tail) - please retry. In case of success it should end with "Collector appears to work".
Also, please check single-threaded case with incremental collection:
gcc -O0 -g -I include -DGC_ASSERTIONS -DGC_PRINT_VERBOSE_STATS tests/test.c extra/gc.c && ./a.out

somasis:
It hit a segmentation fault at the end of the process, I guess it didn't get redirected to the log file.

Incremental+single-threaded: http://ix.io/ntg

ivmai:
No, the log does no correspond to incremental mode (it does not contain "Switched to incremental mode").
I saw it in your first log (http://ix.io/nkP) - can you please list content of config.h?

PR #88, opened 2016-02-01
GC_malloc_explicitly_typed generalization with kind

tholp:
Generalize GC_malloc_explicitly_typed by calling GC_malloc_kind.  This
will also enable thread local allocation of typed memory.

The same kind of generalization should probably be made for
GC_calloc_explicitly_typed and
GC_malloc_explicitly_typed_ignore_off_page, but we don't use these, so
we cannot validate it.

### DIFF
--- a/typd_mlc.c
+++ b/typd_mlc.c
@@ -14,6 +14,7 @@
  */
 
 #include "private/gc_pmark.h"
+#include "gc_inline.h"
 
 /*
  * Some simple primitives for allocation with explicit type information.
@@ -343,10 +344,6 @@ GC_make_sequence_descriptor(complex_descriptor *first,
   }
 #endif
 
-STATIC ptr_t * GC_eobjfreelist = NULL;
-
-STATIC ptr_t * GC_arobjfreelist = NULL;
-
 STATIC mse * GC_typed_mark_proc(word * addr, mse * mark_stack_ptr,
                                 mse * mark_stack_limit, word env);
 
@@ -359,18 +356,14 @@ STATIC void GC_init_explicit_typing(void)
 
     GC_STATIC_ASSERT(sizeof(struct LeafDescriptor) % sizeof(word) == 0);
     /* Set up object kind with simple indirect descriptor. */
-      GC_eobjfreelist = (ptr_t *)GC_new_free_list_inner();
       GC_explicit_kind = GC_new_kind_inner(
-                            (void **)GC_eobjfreelist,
                             (WORDS_TO_BYTES((word)-1) | GC_DS_PER_OBJECT),
                             TRUE, TRUE);
                 /* Descriptors are in the last word of the object. */
       GC_typed_mark_proc_index = GC_new_proc_inner(GC_typed_mark_proc);
     /* Set up object kind with array descriptor. */
-      GC_arobjfreelist = (ptr_t *)GC_new_free_list_inner();
       GC_array_mark_proc_index = GC_new_proc_inner(GC_array_mark_proc);
       GC_array_kind = GC_new_kind_inner(
-                            (void **)GC_arobjfreelist,
                             GC_MAKE_PROC(GC_array_mark_proc_index, 0),
                             FALSE, TRUE);
       for (i = 0; i < WORDSZ/2; i++) {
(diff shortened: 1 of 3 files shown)


typd_mlc.c
```diff
@@ -596,37 +597,17 @@ GC_API GC_descr GC_CALL GC_make_descriptor(const GC_word * bm, size_t len)
 GC_API GC_ATTR_MALLOC void * GC_CALL GC_malloc_explicitly_typed(size_t lb,
                                                                 GC_descr d)
 {
-    ptr_t op;
     size_t lg;
-    DCL_LOCK_STATE;
-
     GC_ASSERT(GC_explicit_typing_initialized);
     lb += TYPD_EXTRA_BYTES;
+    void *result = GC_malloc_kind(lb, GC_explicit_kind);
```
ivmai: if (EXPECT(NULL == result, FALSE)) return NULL;

typd_mlc.c
```diff
@@ -596,37 +597,17 @@ GC_API GC_descr GC_CALL GC_make_descriptor(const GC_word * bm, size_t len)
 GC_API GC_ATTR_MALLOC void * GC_CALL GC_malloc_explicitly_typed(size_t lb,
                                                                 GC_descr d)
 {
-    ptr_t op;
     size_t lg;
-    DCL_LOCK_STATE;
-
     GC_ASSERT(GC_explicit_typing_initialized);
     lb += TYPD_EXTRA_BYTES;
```
ivmai: declaration should precede any statement

ivmai:
I have started some small refactoring and suddenly realized that there is something inconsistent/wrong in the concept. Can you remove GC_eobjfreelist?

tholp:
I have eliminated it and also implemented the change for arrays (but as mentioned we to not use arrays, so I have not been able to test that change).

ivmai:
Please check result of each  GC_malloc_X call to avoid SigSegv

ivmai:
I will consider both of the proposed versions - I don't fully like your 2nd version (e.g. because of API incompatible change) but I don't have immediate answer of how it should look like to be accepted.

ivmai:
I have committed 5e15c58 which prevents creation of kinds that are used for GC_malloc_kind, so the proposed patch (and its previous version) does not work now.
Actually, now I understand that something wrong in the design of _freelists and it is wrong to have THREAD_FREELISTS_KINDS and PREDEFINED_KINDS different from MAXOBJKINDS.
E.g., the patch proposed by you will not work with the current value of PREDEFINED_KINDS, on the opposite legacy code will be broken if PREDEFINED_KINDS is increased.
I think to solve the issue, 
void *_freelists[PREDEFINED_KINDS][MAXOBJGRANULES + 1]
should be changed to
void *(*_freelists[MAXOBJKINDS])[MAXOBJGRANULES + 1]
(similar for thread_local_freelists) - lists for PTRFREE, NORMAL, etc. will be preallocated, others should be allocated explicitly by a client, the change should be transparent to legacy code but newer code, will be able to use GC_malloc_kind and benefit from TLS-based allocations.
However this way, some small slowdown is expected bacause of extra indirection in _freelists.

After that we could return to GC typed allocations improvement.

tholp:
I think that sounds like a good idea.

tholp:
I think that sounds like a good idea. I will close this pull request and reopen another later.

ivmai:
Good. As usual try to make it in small commits. I think first thing is to change _freelists in _GC_arrays and remove PREDEFINED_KINDS without changing _freelists (and THREAD_FREELISTS_KINDS) in struct thread_local_freelists. Details: add _objfreelist, _aobjfreelist, etc. fields back to _GC_arrays (in addition to _freelists field), initialize _freelists[0] to &_objfreelist, etc. (in GC_malloc_kind as GC_arrays should stay in .bss), replace PREDEFINED_KINDS with MAXOBJKINDS (and remove some assertions comparing these 2 macros).

Even more, for your purpose it should be enough because you could compile GC with CFLAGS=-DTHREAD_FREELISTS_KINDS=<increased_value> so that GC_malloc_kind uses TLS-based allocation for custom types automatically.

To avoid potential complains about slight performance degradation of GC_malloc[_atomic] in single-threaded case (because of if-unlikely statement to check for _freelists[0 or 1] is null and 1 extra indirection), we could (in a separate commit) return back the specific implementation of GC_malloc[_atomic] which use _objfreelist and _aobjfreelist directly).

So, you could start preparing 2 commits almost in parallel - refactoring of GC_arrays._freelists and refactoring of GC_malloc_explicitly_typed() (to use GC_malloc_kind).

tholp:
Do I understand you correct if 
    void _freelists[PREDEFINED_KINDS][MAXOBJGRANULES + 1]
 is changed to
    void **_freelists[MAXOBJKINDS];
So that _freelists in itself only contains MAXOBJECTKINDS pointers.

Which then each are either NULL (=unused) or points to a  an array of pointers of size MAXOBJGRANULES + 1

ivmai:
Yes. Correct. But in GC_arrays only for now.
-- пятница, 20 мая 2016г., 16:51 +03:00 от Thomas Linder Puls < notifications@github.com> :

PR #89, opened 2016-02-02
Fix Cygwin64 build

okuoku:
On x86_64, Cygwin symbols do not have leading underscore anymore so we need
an additional underscore when we access linker generated symbols from C.
See https://cygwin.com/ml/cygwin-apps/2013-04/msg00283.html

* include/gc.h (GC_DATASTART, GC_DATAEND): Fix symbol names by adding
underscores on __data_start__, __bss_start__, __data_end__ and __bss_end__.


ivmai:
The patch is ok - I will merge it in a couple of days.
Thank you, actually I found the issue a week ago but had no time to investigate it.

ivmai:
The patch merged to master (with some code reformatting and commit message change), see 4e24d21

PR #91, opened 2016-02-06
Add signal mask to GC marker threads

qykth-git:
When one user thread sets its own signal handler in multi-threaded code, GC marker thread sometimes steal and drop user signal calls. Because GC marker thread doesn't knows the user thread's signal handler, and use default signal handler that ignores any signal call. This becomes to drop the signal call.

This patch inhibits to accept signals by GC marker threads except GC system signals, and don't drop user signal calls.

This patch comes from [Gauche] ( https://github.com/shirok/Gauche ), and works well in long time.

* `pthread_support.c` (`GC_start_mark_threads_inner`): Add signal mask to GC marker threads, and do not to accidentally steal user signal calls.

### DIFF
--- a/pthread_support.c
+++ b/pthread_support.c
@@ -394,6 +394,9 @@ STATIC pthread_t GC_mark_threads[MAX_MARKERS];
 {
     int i;
     pthread_attr_t attr;
+#   ifndef GC_NO_PTHREAD_SIGMASK
+      sigset_t set, oldset;
+#   endif /* !GC_NO_PTHREAD_SIGMASK */
 
     GC_ASSERT(I_DONT_HOLD_LOCK());
     if (available_markers_m1 <= 0) return;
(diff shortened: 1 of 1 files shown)


pthread_support.c
```diff
@@ -424,6 +425,16 @@ STATIC pthread_t GC_mark_threads[MAX_MARKERS];
         }
       }
 #   endif /* HPUX || GC_DGUX386_THREADS */
+
+    /* Apply special signal mask to GC marker threads, and don't drop user */
+    /* defined signals by GC marker threads.                               */
+    sigfillset(&set);
+    sigdelset(&set, GC_get_suspend_signal()    ); /* used by GC to stop the world    */
```
ivmai: Suspend/resume signals are not used for GC_DARWIN_THREADS, GC_OPENBSD_UTHREADS and NACL

pthread_support.c
```diff
@@ -424,6 +425,16 @@ STATIC pthread_t GC_mark_threads[MAX_MARKERS];
...
+    sigdelset(&set, GC_get_thr_restart_signal()); /* used by GC to restart the world */
+    if (pthread_sigmask(SIG_BLOCK, &set, &oldset) < 0) {
+      WARN("pthread_sigmask failed, errno = %" WARN_PRIdPTR "\n", errno);
```
ivmai: In case of pthread_sigmask failure, oldset is uninitialized so the second pthread_sigmask (below) call should not be performed. Also do we really want to start marker thread even if pthread_sigmask failed here? I suggest here to add: GC_markers_m1 = 0; (void)pthread_attr_destroy(&attr); return;

pthread_support.c
```diff
@@ -424,6 +425,16 @@ STATIC pthread_t GC_mark_threads[MAX_MARKERS];
         }
       }
 #   endif /* HPUX || GC_DGUX386_THREADS */
+
+    /* Apply special signal mask to GC marker threads, and don't drop user */
+    /* defined signals by GC marker threads.                               */
+    sigfillset(&set);
```
ivmai: Please do not exceed 78 chars in a line

pthread_support.c
```diff
@@ -394,6 +394,7 @@ STATIC pthread_t GC_mark_threads[MAX_MARKERS];
 {
     int i;
     pthread_attr_t attr;
```
ivmai: pthread_sigmask is not available on some targets, see description of GC_NO_PTHREAD_SIGMASK

ivmai: Other other hand, should we make similar change to win32_threads.c?

qykth-git: I don't know how signals works in MS-Windows.
But if GC marker thread also steals/drops user signals in `win32_thread`, I think similar patch might be required for `win32_thread`.

ivmai:
Please squash all commits into a single one

qykth-git:
Squashed patches into single one.

ivmai:
Committed to master (with some changes). See 95947bd

PR #120, opened 2016-05-23
GC_arrays._freelists increase size to MAXOBJKIND (make actual freelists indirect)

tholp:
The fundamental change is from:
void *_freelists[PREDEFINED_KINDS][MAXOBJGRANULES + 1];
to:
void **_freelists[MAXOBJKINDS];

Individual "PREDEFINED" free lists are allocated as explicit members in
GC_arrays (which is alreadu excluded from static roots).  These are
linked into _freelists in GC_init().


ivmai:
It turned out that the base functionality for alloc generalization existed already long ago, so instead of the proposed commit I have applied this small commit f478a1e and reverted 41871b9.

It should fit your needs, right?

PR #121, opened 2016-06-08
Dump the block information in CSV format

PaulBone:
Dump the block information in CSV format

This makes it easy to copy this information into a new file and process it
with tools that understand CSV, for example to create histograms of block
utilisation.

Thanks


ivmai:
Looks good, I will apply to master branch

ivmai:
Committed to master with some small modifications:  d55c08a

PR #123, opened 2016-06-22
check for execinfo.h directly

vapier:
The current header depends on glibc/uClibc version checks to determine
whether execinfo.h exists which breaks other C libs.  Instead, add an
explicit configure check for it.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -473,6 +473,12 @@ case "$host" in
 esac
 AM_CONDITIONAL(AVOID_CPP_LIB,test $avoid_cpp_lib = yes)
 
+# Check for various headers.
+AC_CHECK_HEADERS([execinfo.h])
+if test "$ac_cv_header_execinfo_h" != "yes"; then
+  AC_DEFINE([GC_MISSING_EXECINFO_H], [1], [Missing execinfo.h header])
+fi
+
 # extra LD Flags which are required for targets
 case "${host}" in
   *-*-darwin*)
--- a/include/gc_config_macros.h
+++ b/include/gc_config_macros.h
@@ -287,7 +287,8 @@
 #   include <features.h>
 # endif
 # if (__GLIBC__ == 2 && __GLIBC_MINOR__ >= 1 || __GLIBC__ > 2) \
-        && !defined(__ia64__) && !defined(__UCLIBC__) \
+        && !defined(__ia64__) \
+        && !defined(GC_MISSING_EXECINFO_H) \
         && !defined(GC_HAVE_BUILTIN_BACKTRACE)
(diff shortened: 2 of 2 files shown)


ivmai:
This patch breaks clients that do not use configure (they would get GC_HAVE_BUILTIN_BACKTRACE never defined by the patch). Is it possible to replace HAVE_EXECINFO_H with HAVE_NO_EXECINFO_H (you may propose a better name, of course) so that to replace !defined(__UCLIBC__) with !defined(HAVE_NO_EXECINFO_H) (thi of course still break some clients but I believe this group is smaller than glibc one).

vapier:
i've added a new symbol `GC_MISSING_EXECINFO_H` that configure will define when execinfo.h is not found.  wdyt?

ivmai:
Better but still to be improved. configure generates config.h which is used only during gc build so the macro specified in config.h should not be used (even indirectly) in include/*.h files except for include/private/*.h ones. (You might find gc_config_macros.h file a complicated one, that's actually because of its functionality is to determine presence of particular features when client includes gc.h.)

vapier:
it's not clear how you want this structured then if `config.h` can't be used.  doesn't seem like any header files which are generated/controlled by configure are installed.

ivmai:
I don't have immediate answer.
> The current header depends on glibc/uClibc version checks to determine whether execinfo.h exists which breaks other C libs
Which ones?

vapier:
`execinfo.h` isn't in POSIX, so all alternative C libs can fail (musl & dietlibc come to mind).  uClibc also has a config option to enable support, so even the current uClibc check isn't entirely correct.

what build systems other than configure+make do you support here ?

ivmai:
CMake is also used. Plus some clients (like me) build it from command line, e.g.: gcc -I include tests/test.c extra/gc.c

ivmai:
Please look at GC_HAVE_BUILTIN_BACKTRACE in configure.ac

vapier:
but that's how i'm doing things now -- using `AC_DEFINE` in configure.ac.  i suspect that usage is broken the same way.  if you run configure on ia64 linux and it defines `GC_HAVE_BUILTIN_BACKTRACE`, the installed headers still won't define `GC_HAVE_BUILTIN_BACKTRACE`.

the only way to have the installed headers work is to define a new one that'll be installed & included and be hand maintained (so we don't get all of the defines/templates that autoheader would insert on us).

ivmai:
Agree. I will look at it again. 
-- пятница, 24 июня 2016г., 10:11 +03:00 от Mike Frysinger  notifications@github.com :

ivmai:
Merged to master - see 978c9b9

PR #126, opened 2016-07-03
Thread local allocation of typed objects

tholp:


tholp:
This will make typed (and typed array) object allocation use the thread local feature.

(The request includes "Missing voilatile declaration")

include/private/gc_priv.h
```diff
@@ -1641,7 +1641,7 @@ GC_EXTERN void (*GC_push_typed_structures)(void);
                         /* the typed allocation support if unused.      */
 
 GC_INNER void GC_with_callee_saves_pushed(void (*fn)(ptr_t, void *),
-                                          ptr_t arg);
+                                          volatile ptr_t arg);
```
ivmai: Why do we need volatile here? Is there a compiler complaining about it? (Even the definition has it, it is meaningless in the declaration

tholp: Yes, Microsoft Visual C++ (with Warning Level 4) gives a warning.

In addition we use "Treat Warnings as Errors" so for us it is an error.

ivmai: OK, merged to master - 65b50c0

PR #127, opened 2016-07-05
Fix assert in mark.c for non-heap regions

GaborDrescher:
Just the changes proposed by @ivmai


ivmai:
Thank you. Merged to master - da2fcda

PR #129, opened 2016-07-07
Typed allocation using kind operations, …

tholp:
… to facilitate usage of thread local storage


ivmai:
Committed to master with some modifications, see 55d6247

PR #130, opened 2016-07-15
PT_GNU_RELRO: Don't warn about not finding PT_GNU_RELRO segment when …

kmatheussen:
…we have defined a GC_has_static_roots_func callback.

It's most likely not found just because the segment had been excluded.

Alternatively, we could have registered all segments, and checked the callback afterwards, but then we could break
programs that rely on GC_has_static_roots_func to avoid overflowing the maximum number of roots. In addition,
it would make the logic slightly more complicated, probably without a very good reason since the chance of this warning
to show without the segment being excluded is likely to be none.

### DIFF
--- a/dyn_load.c
+++ b/dyn_load.c
@@ -490,7 +490,6 @@ STATIC int GC_register_dynlib_callback(struct dl_phdr_info * info,
         continue;
       
 #     ifdef PT_GNU_RELRO
-        if (n_load_segs >= MAX_LOAD_SEGS) ABORT("Too many PT_LOAD segs");
 #       if CPP_WORDSZ == 64
           /* FIXME: GC_push_all eventually does the correct         */
           /* rounding to the next multiple of ALIGNMENT, so, most   */
@@ -499,11 +498,16 @@ STATIC int GC_register_dynlib_callback(struct dl_phdr_info * info,
           /* start pointer value may require aligning */
           start = (ptr_t)((word)start & ~(sizeof(word) - 1));
 #       endif
-        load_segs[n_load_segs].start = start;
-        load_segs[n_load_segs].end = end;
-        load_segs[n_load_segs].start2 = 0;
-        load_segs[n_load_segs].end2 = 0;
-        ++n_load_segs;
+        if (n_load_segs >= MAX_LOAD_SEGS) {
+          WARN("Too many PT_LOAD segs\n", 0);
+          GC_add_roots_inner(start, end, TRUE);
+        } else {
+          load_segs[n_load_segs].start = start;
+          load_segs[n_load_segs].end = end;
+          load_segs[n_load_segs].start2 = 0;
+          load_segs[n_load_segs].end2 = 0;
+          ++n_load_segs;
+        }
 #     else
         GC_add_roots_inner(start, end, TRUE);
 #     endif /* PT_GNU_RELRO */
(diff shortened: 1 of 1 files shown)


ivmai:
2 of 3 commits merged to master - see 1dee912 and 017bd0a
I will review the 3rd one next week.

dyn_load.c
```diff
@@ -480,18 +480,56 @@ STATIC int GC_register_dynlib_callback(struct dl_phdr_info * info,
...
   p = info->dlpi_phdr;
   for( i = 0; i < (int)info->dlpi_phnum; i++, p++ ) {
-    switch( p->p_type ) {
+    if( p->p_type == PT_LOAD) {
+      GC_has_static_roots_func callback = GC_has_static_roots;
+      if( !(p->p_flags & PF_W) ) continue;
+      start = ((char *)(p->p_vaddr)) + info->dlpi_addr;
+      end = start + p->p_memsz;
+
+      if (callback != 0 && callback(info->dlpi_name, start, p->p_memsz))
```
ivmai: It should be: callback != 0 && !callback(...) right?

kmatheussen: Goodness. Yes, that have to be !callback. Fixed.

kmatheussen: Sorry, I should have tested this before commiting. I've tested now, and when using "callback" instead of "!callback", my program crashes almost immediately.

ivmai:
The last one merged to master (with some non-functional modifications), see b5d5c43

dyn_load.c
```diff
@@ -480,18 +480,56 @@ STATIC int GC_register_dynlib_callback(struct dl_phdr_info * info,
...
+          /* FIXME: GC_push_all eventually does the correct         */
+          /* rounding to the next multiple of ALIGNMENT, so, most   */
+          /* probably, we should remove the corresponding assertion */
+          /* check in GC_add_roots_inner along with this code line. */
+          /* start pointer value may require aligning */
+          start = (ptr_t)((word)start & ~(sizeof(word) - 1));
+#       endif
+        if (n_load_segs >= MAX_LOAD_SEGS) {
+          WARN("Too many PT_LOAD segs\n", 0);
+          GC_add_roots_inner(start, end, TRUE);
```
ivmai: I have a question: what will happen when the dynamic library (registered this way) is unloaded (dlclose)?

kmatheussen: I'm not sure what you think about, but I would think, with my limited knowledge about bdwgc, that all roots added with GC_add_roots_inner are cleared after collecting. So I guess things could go really wrong if a library is unloaded while collecting garbage, but not if it is unloaded inbetween two collections.

ivmai:
I see, you are right.
BTW. I have added a small commit preventing repeated warning about load_segs overflow - 33fc2f8

kmatheussen:
Seems like it's relatively easy to avoid this problem though. From the man page of dlopen:

"If  the  same  library  is loaded again with dlopen(), the same file handle is returned.  The dl library maintains reference counts for library handles, so a dynamic library is not deallocated until dlclose() has been called on it as many times as dlopen() has succeeded on it."

So the only thing we have to do is call dlopen()/dlclose() on all opened libraries (if the list of filenames are available, haven't checked) in GC_register_dynlib_callback.

But I guess it's not very important. If there are anyone who calls dlopen()/dlclose() from other threads, they probably call GC_disable() first.

ivmai:
>But I guess it's not very important. If there are anyone who calls dlclose() from other threads, they probably call GC_disable() first.
I agree. And as the disclaimer of dyn_load.c says:
 * None of this is safe with dlclose and incremental collection.
 * But then not much of anything is safe in the presence of dlclose.

PR #139, opened 2016-12-30
Utilisation

PaulBone:
These changes make it easier to analyse heap utilisation.  They make the GC_dump() output easier to parse, and include a python script that can generate tables and graphs. Potential users will want to edit the script so that it can work for their own purposes.

For a good example see http://paul.bone.id.au/2016/12/29/more-about-memory-fragmentation/

Thanks.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -190,6 +190,9 @@ EXTRA_DIST += tools/add_gc_prefix.c tools/gcname.c tools/if_mach.c \
     extra/Mac_files/MacOS_config.h \
     include/private/msvc_dbg.h extra/msvc_dbg.c tools/callprocs.sh
 
+# Extra 'contrib' scripts
+EXTRA_DIST += extra/analyse_dump.py
+
 #
 # :GOTCHA: GNU make rule for making .s out of .S is flawed,
 # it will not remove dest if building fails
--- a/alloc.c
+++ b/alloc.c
@@ -63,12 +63,17 @@ word GC_non_gc_bytes = 0;  /* Number of bytes not intended to be collected */
 
 word GC_gc_no = 0;
 
-#ifndef SMALL_CONFIG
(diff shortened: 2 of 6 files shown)


ivmai:
I haven't reviewed the diff fully but I see some improvements are needed:
* if new file is added then it should be mentioned by EXTRA_DIST in Makefiles (otherwise make dist would produce a tarball w/o the file)
* the change could break compilation for some clients - it is better to add new API function (instead of changing the prototype of the existing public one)
* use TRUE/FALSE in initialization/updating of GC_bool variables
* export getters/setters instead of variables
* if a var (GC_init_time) is used in one file, it is better to define it as STATIC
* it seems that the defined 3 global vars are never used in case of SMALL_CONFIG or NO_DEBUGGING

PaulBone:
Hi @ivmai I've addressed those issues.  Thanks.

ivmai:
Ok, I will process these commits (one-by-one) in 1-2 weeks.

ivmai:
Sorry, I'm still processing it. For now I've merged only the commit about named dumps - e81e5b8

PaulBone:
No problem, We don't need this integrated right away. it just seems like a good idea.

ivmai:
I have refactored the original patches. The final patches merged to master:
* e81e5b8 - named dumps
* c368a45 - print time since GC init in GC_dump
* d0f71f1- accumulation of full collection time

The only thing I have not reviewed and merged is analyse_dump.py. Please update the script to reflect the changes in changes in the above patches, and resubmit it.

PaulBone:
I'm not actively using this code or Boehm GC anymore.  You're welcome to use my code and update the script.  The Mercury team (CC @wangp) may be interested also.

Thanks.

ivmai:
Closing it. Converted to issue #360

PR #141, opened 2017-01-10
Avoiding conditional directives that split parts of statements

RomeroMalaquias:
Hello, this change is only aesthetic, it's a suggestion to avoid the use of conditional directives that break statements i think that kind of concern improve the code maintainability do you agree?

### DIFF
--- a/misc.c
+++ b/misc.c
@@ -1749,6 +1749,7 @@ GC_API GC_warn_proc GC_CALL GC_get_warn_proc(void)
   STATIC void GC_CALLBACK GC_default_on_abort(const char *msg)
   {
     GC_find_leak = FALSE; /* disable at-exit GC_gcollect()  */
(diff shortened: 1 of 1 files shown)


ivmai:
There are several issues with the proposed code:
* the pattern is used in many places in gc (and libatomic_ops)
* 2 code defects (or false positives) - "condition is always true" and/or "value is set but never used" could be reported by compilers and static analysis tools which are not smart enough (e.g. I periodically check the code for defects with gcc/clang -Wpedantic, CSA, cppcheck, coverity, etc.)
* the approach leads to the need to define extra variables (as in your code) or to code duplication
* currently I don't see how could it make code maintainability better

RomeroMalaquias:
Ok, that makes sense.
Thanks for your time.

PR #153, opened 2017-03-06
Increase MAX_HEAP_SECTS

tholp:


ivmai:
Merged to master (see a17c62c). All tests pass (including --enable-large-config [--enable-munmap] for linux/x64 and linux/x86).

PR #164, opened 2017-05-28
Fix uClibc build

bkuhls:
uClibc defines __GLIBC__ but does not contain libc-version.h leading to
a build error:

```
./include/private/gcconfig.h:1426:72: fatal error: gnu/libc-version.h: No such file or directory
 #         include <gnu/libc-version.h> /* for gnu_get_libc_version() */
```


ivmai:
Merged to master: 047230b

PR #201, opened 2018-02-04
Extern "C" around declarations, not includes.

jaykrell:
Leave each header to decide if its content is extern "C" itself.
Except for very old C++-unaware systems that do not use extern "C", each header should own the C-ness of its declarations, and not have its clients determine it.

### DIFF
--- a/include/gc_inline.h
+++ b/include/gc_inline.h
@@ -47,6 +47,10 @@
 # endif
 #endif
 
+#ifdef __cplusplus
+extern "C" {
+#endif
+
 #ifndef GC_PREFETCH_FOR_WRITE
 # if GC_GNUC_PREREQ(3, 0) && !defined(GC_NO_PREFETCH_FOR_WRITE)
 #   define GC_PREFETCH_FOR_WRITE(x) __builtin_prefetch((x), 1)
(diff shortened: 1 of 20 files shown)


jaykrell:
Is a1aa03d implicitly rejected, or just taking more time, or want it restated differently? I've always avoided wrapping other people's headers in extern C, keeping it on my own declarations, that I will decide if I implement them in C or C++. This seems like the right approach. The only flip side I can think of is ancient systems that are lacking extern "C" but need it. Wrapping them can be seen as a favor. As well, perhaps not all of Boehm needed this -- we call just a few functions ourselves.

ivmai:
> Is a1aa03d implicitly rejected, or just taking more time

Taking more time.

jaykrell:
Great. This is the impetus:
https://github.com/mono/mono/pull/6695
but notice I got by with much less.

Specifically part of this:
https://github.com/mono/mono/pull/6695/commits/2ca03a7f4d59cee522b808e80030306074d5677c

and see also part of this:
https://github.com/mono/mono/pull/6695/commits/eb1bf3d7d5e94524f2ca7c38bc511e1825eef5b0

but I acknowledge that void* is not K&R so you might prefer more nuancing, if you support K&R.

Thank you.

jaykrell:
Err I meant https://github.com/mono/mono/pull/6695/commits/7b0cd1e4edff6a0d9812cb186c69d08c41dccb18

ivmai:
Done.

Issue #202, opened 2018-02-06
make check fails on alpine linux s390x


ncopa:
gctest fails on Alpine Linux s390x:

````
...
make  check-TESTS
make[2]: Entering directory '/home/ncopa/aports/main/gc/src/gc-7.6.4'
make[3]: Entering directory '/home/ncopa/aports/main/gc/src/gc-7.6.4'
PASS: cordtest
FAIL: gctest                    
PASS: leaktest                
PASS: middletest
PASS: smashtest                                
PASS: hugetest
PASS: realloc_test
PASS: staticrootstest
PASS: threadleaktest
PASS: threadkey_test
PASS: subthreadcreate_test
PASS: initsecondarythread_test
PASS: test_cpp
PASS: disclaim_test
PASS: disclaim_bench
============================================================================
Testsuite summary for gc 7.6.4
============================================================================
# TOTAL: 15
# PASS:  14
# SKIP:  0
# XFAIL: 0
# FAIL:  1
# XPASS: 0
# ERROR: 0
============================================================================
See ./test-suite.log
Please report to https://github.com/ivmai/bdwgc/issues
============================================================================
make[3]: *** [Makefile:1855: test-suite.log] Error 1
make[3]: Leaving directory '/home/ncopa/aports/main/gc/src/gc-7.6.4'
make[2]: *** [Makefile:1963: check-TESTS] Error 2
make[2]: Leaving directory '/home/ncopa/aports/main/gc/src/gc-7.6.4'
make[1]: *** [Makefile:2297: check-am] Error 2
make[1]: Leaving directory '/home/ncopa/aports/main/gc/src/gc-7.6.4'
make: *** [Makefile:1740: check-recursive] Error 1
ncdev-edge-s390x:~/aports/main/gc/src/gc-7.6.4$ cat test-suite.log 
================================
   gc 7.6.4: ./test-suite.log
================================

# TOTAL: 15
# PASS:  14
# SKIP:  0
# XFAIL: 0
# FAIL:  1
# XPASS: 0
# ERROR: 0

.. contents:: :depth: 2

FAIL: gctest
============

Segmentation fault (core dumped)
FAIL gctest (exit status: 139)

````

ncopa:
I think it runs out of stack space.

ivmai:
1. What's the value of BIG macro? (defined in tests/test.c)
2. Is NO_GETCONTEXT macro defined in include/private/gcconfig.h?

ncopa:
[...]

PR #203, opened 2018-02-06
Fix gctest with musl libc on s390x

ncopa:
Issue #202 (bdwgc).

DEFAULT_STACK_MAYBE_SMALL is only set if PARALLEL_MARK is set, but stack
is too small on musl libc even if PARALLEL_MARK is unset. We solve this
by detecting musl via NO_GETCONTEXT.


ivmai:
Build failures are not relevant to this commit.

ivmai:
Merged f8a6bd7

PR #204, opened 2018-02-09
Receiving void* instead of char* is easier to use, as it requires no casting.

jaykrell:
Small downside is anyone using decltype(GC_push) in C++ function signatures
will get different name mangling.

### DIFF
--- a/include/gc_mark.h
+++ b/include/gc_mark.h
@@ -281,9 +281,9 @@ GC_API void GC_CALL GC_set_mark_bit(const void *) GC_ATTR_NONNULL(1);
 /* (GC_push_conditional pushes either all or only dirty pages depending */
 /* on the third argument.)  GC_push_all_eager also ensures that stack   */
 /* is scanned immediately, not just scheduled for scanning.             */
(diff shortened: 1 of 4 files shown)


ivmai:
Please fix build failures, and also revise all GC_push_all/all_eager/conditional calls to remove unneeded casts.

jaykrell:
The casts are ok, but now optional, esp. in clients. But ok, and definitely failures.

jaykrell:
Hm, so volatile x* doesn't go silently to void*, either a) cast sometimes, when there is volatile, which will be rare, or b) change the parameters to volatile void*, which is my very slight preference.

ivmai:
> Hm, so volatile x* doesn't go silently to void*, either a) cast sometimes, ...

Yes, like done in GC_clear_stack_inner() (in misc.c) for a BZERO/memset call:
      BZERO((/* no volatile */ void *)dummy, sizeof(dummy));

ivmai:
Merged (with some changes) as b2b0d27

PR #208, opened 2018-02-18
Add RISC-V support.

shlevy:
Requires https://github.com/ivmai/libatomic_ops/pull/32

### DIFF
--- a/include/private/gcconfig.h
+++ b/include/private/gcconfig.h
@@ -650,6 +650,15 @@
 #   endif
 #   define mach_type_known
 # endif
+# if defined(__riscv) && defined(LINUX)
+#   if __riscv_xlen == 32
+#     define RISCV32
+#     define mach_type_known
+#   elif __riscv_xlen == 64
+#     define RISCV64
+#     define mach_type_known
+#   endif
+# endif
 
 # if defined(SN_TARGET_PSP2)
 #   define mach_type_known
@@ -2961,6 +2970,32 @@
 #   endif
 #   define PREFETCH(x) __insn_prefetch_l1(x)
 #   define CACHE_LINE_SIZE 64
+#   ifdef LINUX
+#     define OS_TYPE "LINUX"
+      extern int __data_start[];
+#     define DATASTART ((ptr_t)__data_start)
+#     define LINUX_STACKBOTTOM
+#     define DYNAMIC_LOADING
+#   endif
+# endif
+
+# ifdef RISCV32
+#   define CPP_WORDSZ 32
+#   define MACH_TYPE "RISC-V 32"
+#   define ALIGNMENT 4
+#   ifdef LINUX
+#     define OS_TYPE "LINUX"
+      extern int __data_start[];
+#     define DATASTART ((ptr_t)__data_start)
+#     define LINUX_STACKBOTTOM
+#     define DYNAMIC_LOADING
+#   endif
+# endif
+
+# ifdef RISCV64
+#   define CPP_WORDSZ 64
+#   define MACH_TYPE "RISC-V 64"
+#   define ALIGNMENT 8
 #   ifdef LINUX
 #     define OS_TYPE "LINUX"
       extern int __data_start[];


shlevy:
Currently only compile-tested on risc64, will report back when runtime tested. Thanks to @sorear for the pointers.

shlevy:
I've now done some light runtime testing (via the repl for the nix expression language) and all seems good so far. Do you have a stress test I can run?

ivmai:
> Requires ivmai/libatomic_ops#32

No, it does not. The code in master branch relies on GCC built-in primitives if test_atomic_ops passes.

brucehoult:
@palmer-dabbelt would you have a chance to test Boehm GC native on Hifive Unleashed? Should take well under five minutes of your time.

```
git clone git://github.com/ivmai/bdwgc.git
cd bdwgc
./autogen.sh
./configure
make -j4
make check
```

ivmai:
After merging your patch, I found another variant of the patch (https://github.com/AOSC-Dev/aosc-os-abbs/blob/staging/base-libs/gc/autobuild/patches/gc-riscv.patch) created by @Icenowy (the difference is in DATAEND and DYNAMIC_LOADING).

shlevy:
@ivmai I will verify that DYNAMIC_LOADING works here. How can I determine which DATAEND is correct?

shlevy:
Ah, I'll do a native build and run the tests when I get my native toolchain bootstrapped.

ivmai:
> the difference is in DATAEND and DYNAMIC_LOADING

I'm sorry, no difference (DYNAMIC_LOADING exists in both patches, DATA is defined to _end by default).
So your patch is OK, I've merged it already. The only thing, I think, it would be nice to collapse RISCV32/64 cases into single RISCV (like it is done e.g. for MIPS which could be both 32 and 64-bit too).

shlevy:
@ivmai OK, I'll do that, thanks!

shlevy:
Merged in 4f7f0eebd24dcde9f2b3ec2cb98913fc39bbdda3

palmer-dabbelt:
@brucehoult Can you point me to a disk image (which you could generate with QEMU) that has all the stuff you want me to run, and then I can run it on the board?  I can't figure out how to use Fedora :)

brucehoult:
@palmer-dabbelt that's exactly why I pinged you .. it's tooooo hard to do with cross compilers and qemu and just busybox (especially given this stuff is just in my copious free time), but I figured you'd have a real GNU userland and toolchain on your board by now :-)

It's not exactly urgent,. It would be good to have once people start to have real boards in their hands, but once we have real boards in our hands we can do it easily. I sounds as if Fedora will be there by the end of next month.

Did I get the correct impression that one of the *BSDs is done and ready to go?

palmer-dabbelt:
Fedora should have native toolchain packages now... :)

rwmjones:
Fedora image is here: https://fedorapeople.org/groups/risc-v/disk-images/ and indeed it is true that we have a native toolchain based on gcc 7.3.1, binutils 2.30 and glibc 2.27.  You need to read the readme file in that link to understand how to use it.

I am running the test above using the latest version from git (f9c5815c).  Will report back once it has finished.

rwmjones:
Note these results are on qemu, not on real hardware.  I used `-smp 4` to emulate 4 CPUs.

```
PASS: cordtest
./test-driver: line 107:  4801 Aborted                 (core dumped) "$@" > $log_file 2>&1
FAIL: gctest
PASS: leaktest
PASS: middletest
PASS: smashtest
PASS: hugetest
PASS: realloc_test
PASS: staticrootstest
PASS: test_atomic_ops
PASS: threadleaktest
PASS: threadkey_test
PASS: subthreadcreate_test
PASS: initsecondarythread_test
PASS: disclaim_test
PASS: disclaim_bench
============================================================================
Testsuite summary for gc 7.7.0
============================================================================
# TOTAL: 15
# PASS:  14
# SKIP:  0
# XFAIL: 0
# FAIL:  1
# XPASS: 0
# ERROR: 0
============================================================================
See ./test-suite.log
Please report to https://github.com/ivmai/bdwgc/issues
============================================================================
```

I'll see if I can get a stack trace from the failure.

rwmjones:
I just realised we don't have gdb, and we're some way from having it, so no stack trace for a while ...  If there's anything else that can be done to diagnose problems (verbose mode?) let me know.

PR #220, opened 2018-05-14
add redox to config script

vmchale:
This doesn't add full support for threads on redox but it does stop the config script from breaking.


ivmai:
Is redox already supported in gcconfig.h?

ivmai:
> This doesn't add full support for threads

What do you mean by "full support"? 

PS. If the collector has no threads support for the target yet then it could be built for the target passing --disable-threads to configure, (The patch is not needed in this case.)

PR #232, opened 2018-08-28
mach/iOS12 fix

### LINKED ISSUE #231: Darwin/ios12 -- GC_start_world doesn't resume all threads
This happens due mach thread ports are released in GC_stop_world as result in some cases system creates new and GC_start_world will not find thread to resume. as result threads will stay suspended. When this happen to GDC threads app is terminated as message loop is halt

dkimitsa:
this PR fixes #231
* thread mach port is not deallocated when thread is saved in GC_mach_threads as it cause thread not being resumed as thread port object can be changed
* threads are now being suspended without check of suspend_count (as it can be suspended by app)
* changed logic of resume, now is primary cycle list are threads in GC_mach_threads

investigation and tech details are in series of posts(to long read as for comment) https://dkimitsa.github.io/2018/08/19/investigating-ios12-beta-crash-vol3/

### DIFF
--- a/darwin_stop_world.c
+++ b/darwin_stop_world.c
@@ -441,7 +441,7 @@ GC_INNER void GC_push_all_stacks(void)
 
   struct GC_mach_thread {
     thread_act_t thread;
-    GC_bool already_suspended;
+    GC_bool suspended;
   };
 
   struct GC_mach_thread GC_mach_threads[GC_MAX_MACH_THREADS];
@@ -451,7 +451,8 @@ GC_INNER void GC_push_all_stacks(void)
 /* returns true if there's a thread in act_list that wasn't in old_list */
 STATIC GC_bool GC_suspend_thread_list(thread_act_array_t act_list, int count,
                                       thread_act_array_t old_list,
-                                      int old_count, mach_port_t my_thread)
+                                      int old_count, task_t my_task,
+                                      mach_port_t my_thread)
 {
   int i;
   int j = -1;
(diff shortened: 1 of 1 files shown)


darwin_stop_world.c
```diff
@@ -684,8 +663,10 @@ GC_INLINE void GC_thread_resume(thread_act_t thread)
 # endif
   /* Resume the thread */
   kern_result = thread_resume(thread);
+# ifdef DEBUG_THREADS
   if (kern_result != KERN_SUCCESS)
-    ABORT("thread_resume failed");
+    GC_log_printf("thread_resume failed %p\n", (void *)thread);
```
ivmai: what are the cases when we should ignore thread_resume failure? can it happen when GC_thread_resume is called having GC_query_task_threads is false?

dkimitsa: during having stress test on this I faced the issue when it failed to resume previously successfully suspended thread. it is rare case I met it like several times a day. Had no clear understanding why it was happening under iOS12. Error code usually was "mach port invalid"
considering that current logic allows failure to suspend thread as normal case -- it might be ok to allow application survive when there is thread resume failure as well.

ivmai: > Error code usually was "mach port invalid"
KERN_INVALID_ARGUMENT error code? Any other error code?

ivmai: 1. Have you tested the path in 2 variants:  CFLAGS_EXTRA=-DGC_NO_THREADS_DISCOVERY and CFLAGS_EXTRA=-DGC_DISCOVER_TASK_THREADS.

2. What is the value of GC_query_task_threads when you meat resume failure? Is it always false, or  always true, or any?

dkimitsa: hi @ivmai 
I played a little with stress test today to get exact error number but was not able to reproduce on recent ios12 beta. will check later today and come back with result. 

> Have you tested the path in 2 variants: CFLAGS_EXTRA=-DGC_NO_THREADS_DISCOVERY and CFLAGS_EXTRA=-DGC_DISCOVER_TASK_THREADS.

very first "fix" was to use `--disable-threads-discovery` at my end. It is not big surprise it worked as each managed thread in this case keeps own thread's match port (e.g. no dealloc happening)

> What is the value of GC_query_task_threads when you meat resume failure? Is it always false, or always true, or any?

Sorry I have no this information as I'm stress testing only isolated stop/start world code snippets in very empty Xcode project without bdwgc itself.

ivmai:
Thank you. Committed with some code refactoring as 59c82a5

Issue #237, opened 2018-09-10
Race between GC_remove_counts and GC_find_header


ivmai:
Build link: https://travis-ci.org/ivmai/bdwgc/jobs/436513916

Source: master (or 8.0.0)
Host: Linux/x64
Reproduction frequency on Travis CI is high (~1/10).

How build and run: ./configure --enable-gc-assertions --enable-gc-debug --enable-handle-fork=manual --enable-large-config --with-libatomic-ops=no && make check CFLAGS_EXTRA="-fsanitize=thread -D NO_CANCEL_SAFE -D NO_INCREMENTAL -D USE_SPIN_LOCK -fno-omit-frame-pointer -D TEST_FORK_WITHOUT_ATFORK"

Output (of gctest):

WARNING: ThreadSanitizer: data race (pid=11460)
  Write of size 8 at 0x7ff4dd737000 by thread T20:
    # 0 GC_remove_counts /home/travis/build/ivmai/bdwgc/extra/../headers.c:312:9 (libgc.so.1+0x167b7)
    # 1 GC_freehblk /home/travis/build/ivmai/bdwgc/extra/../allchblk.c:870 (libgc.so.1+0x167b7)
    # 2 GC_add_to_heap /home/travis/build/ivmai/bdwgc/extra/../alloc.c:1260:5 (libgc.so.1+0x1e159)
    # 3 GC_expand_hp_inner /home/travis/build/ivmai/bdwgc/extra/../alloc.c:1386:5 (libgc.so.1+0x1e7dd)
    # 4 GC_collect_or_expand /home/travis/build/ivmai/bdwgc/extra/../alloc.c:1482:10 (libgc.so.1+0x1ec26)
    # 5 GC_allocobj /home/travis/build/ivmai/bdwgc/extra/../alloc.c:1545:18 (libgc.so.1+0x1f02e)
    # 6 GC_generic_malloc_inner /home/travis/build/ivmai/bdwgc/extra/../malloc.c:191:18 (libgc.so.1+0xd14a)
    # 7 GC_generic_malloc_many /home/travis/build/ivmai/bdwgc/extra/../mallocx.c:473:12 (libgc.so.1+0x2e411)
    # 8 GC_malloc_kind /home/travis/build/ivmai/bdwgc/extra/../thread_local_alloc.c:179:5 (libgc.so.1+0x2c137)
[...]

PR #239, opened 2018-10-08
Add test using disclaim notifiers to implement a weak map.

paurkedal:
This test is more in the spirit of my original motivation for the disclaim notifiers, which relies on returning non-zero to block collection of objects.  My old library code is highly optimised and too involved to lift into the test, so I wrote a hash map from scratch.  It took a bit more time than expected to debug, but the result is fairly straight forward.  Note that the "Skipping marked..." case can happen, it's just extremely rare.


paurkedal:
Yes, I noticed this need more work. Updating now with at least the warnings fixed, and I think the uninitialized var was fixed by zeroing a struct which probably contained some padding. We'll see in a moment.

I am not sure what the last test failure is, as I don't have a working build environment (the instrumentation gives segfault in all tests).  I may need to revisit this later if I don't figure it out this week.

A am not sure merging everything into `disclaim_test.c` is wise.  Both tests are heavy and we have seen in the past that it may take dozens of re-runs to reproduce an error.  Also I think it is useful to see whether the two tests fail independently, in case they fail consistently.

paurkedal:
I found the print statements in `.travis.yml`, so there should be output from the failing tests now.

ivmai:
> Also I think it is useful to see whether the two tests fail independently, in case they fail consistently.

OK.

paurkedal:
I made a minor optimization to prevent CodeFactor from discovering the duplication.  If we want a real fix, I can do that e.g. by defining the loop code as a higher order function and pass it in a shared source file linked to both tests.

Otherwise I believe this should pass now.

paurkedal:
The remaining two test failures occur in `gctest`, so must be unrelated to this PR.

ivmai:
> I made a minor optimization to prevent CodeFactor from discovering the duplication.

OK.

> The remaining two test failures occur in gctest, so must be unrelated to this PR.

Yes, I've met both them in the past (#237 and #240) but I haven't figured out how to provide the correct synchronization yet.

ivmai:
>  Note that the "Skipping marked..." case can happen, it's just extremely rare.

Is it OK that I also see always "0 removed, 0 locked"?

paurkedal:
On Fri, 19 Oct 2018 at 08:49, Ivan Maidanski <notifications@github.com>
wrote:

> Note that the "Skipping marked..." case can happen, it's just extremely
> rare.
>
> Is it OK that I also see always "0 removed, 0 locked"?
>

No, then something is wrong, probably the disclaim procedure not invoked.
Under which configuration did you see that?

paurkedal:
Is also the fin. ratio of disclaim_test is zero?

ivmai:
> Is also the fin. ratio of disclaim_test is zero?

disclaim_bench output on Cygwin/x64:
```
                         fin. ratio       time/s    time/fin.
regular finalization:       0.9922        3.719   8.9366e-07
 finalize on reclaim:       0.9999        0.656  1.56413e-07
     no finalization:       0.0000        0.328          N/A
```

ivmai:
I've done some code refactoring/reformatting and enabling it for single-threaded builds (6eba421).

However, today I've found there's some issue when incremental collection is on (both manual or mprotect-based). The failures: https://travis-ci.org/ivmai/bdwgc/jobs/444577295, https://travis-ci.org/ivmai/bdwgc/jobs/444577300:
Assertion failure: extra/../reclaim.c:437

How to reproduce it locally (linux/x64):
./configure --disable-parallel-mark --disable-thread-local-alloc --enable-gc-assertions && GC_ENABLE_INCREMENTAL=1 make -j check

ivmai:
There's also a strange data race (https://travis-ci.org/ivmai/bdwgc/jobs/444577330):
TSan reports it for a variable modified while holding wm lock.

paurkedal:
Thanks for investigating.  I'm busy at a conference this week, but will try to help out when I return.

ivmai:
> The stat_* functions need to be atomically updated, since the locking of the map is granular (cf WEAKMAP_MUTEX_COUNT) to allow parallelism between different key hashes.

Right, not strange, thank you. I'll fix it.

Issue #252, opened 2018-11-21
disclaim_weakmap_test fails if FIND_LEAK


ivmai:
Source: latest master
Host: Ubuntu/x64
How to reproduce: ./configure --enable-werror --enable-gc-assertions --disable-parallel-mark --disable-thread-local-alloc && make -j check CFLAGS_EXTRA="-D FIND_LEAK"

The error in disclaim_weakmap_test.log is either
```Assertion failure, line 407: pair_new(p0, p1) == pop[t]```
Or
```pthread_mutex_trylock: Invalid argument```

The issue is also reported in #250.
The test was added in pull #239.

\+ @paurkedal

paurkedal:
The assertion failure is most likely due to `GC_print_all_errors` calling `GC_free` on objects which should have been kept according to the disclaim callback.  I haven't though about how to deal with it yet.

The `EINVAL` issue is more obscure.  The failures seem to happen only from the main thread, while subthreads are happily locking and unlocking.  According the the `pthread_mutex_lock` POSIX man page, the cause of `EINVAL` is related to thread priority when using the `PTHREAD_PRIO_PROTECT` mutex protocol attribute.  However, the default is `PTHREAD_PRIO_NONE`, and we're not setting it.  The man page also suggests that `EINVAL` may be returned if the argument is invalid, as one might expect.  But if that was the case, the calls in subthreads should have failed, as well.

ivmai:
The issue is caused by GC_gcollect invoked at-exit (after weakmap is destroyed). It's easy to fix, I'll do it.

ivmai:
Fixed by commit 20cfa75

ivmai:
It looks like it is not fixed completely:
https://travis-ci.org/ivmai/bdwgc/jobs/458917783 failed with:
Checksum failure for 0x2af5d4fff1e8 = (0x2af5d4fff1e8, 0x2af5d4fdd6c8) at 413

paurkedal:
Also I still get the assertion failure occasionally, so yes, seems like only the pthread error was solved.  I think the cause for the remaining errors are more clear though:

`GC_free` on objects from the weak map will invalidate objects referring to them, and this is what happens in leak detection mode, as the disclaim notifier is not consulted.  This can cause two identically hash-constructed objects ending up two different live pointers, as directly apparent from the assert failure.  It can also cause the checksum failure of a referrer when the referred object is re-initialized after `GC_free`.

[...]

PR #253, opened 2018-11-26
Disclaim callbacks are incompatible with leak detection mode.

paurkedal:
This is my proposal for #252.

* fnlz_mlc.c (GC_finalized_malloc): Return NULL if GC_find_leak.
* include/gc_disclaim.h: Document incompatibility with leak detection
  mode and GC_free.
* tests/disclaim_bench.c: Skip for leak detection mode.
* tests/disclaim_test.c: Likewise.
* tests/disclaim_weakmap_test.c: Likewise.

### DIFF
--- a/fnlz_mlc.c
+++ b/fnlz_mlc.c
@@ -94,6 +94,8 @@ GC_API GC_ATTR_MALLOC void * GC_CALL GC_finalized_malloc(size_t lb,
 {
     word *op;
 
+    if (EXPECT(GC_find_leak, FALSE))
+        return NULL;
     GC_ASSERT(GC_finalized_kind != 0);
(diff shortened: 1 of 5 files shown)


ivmai:
I fixed it in a different way (maintaining consistency with GC_register_finalizer for find-leak mode) - see my comment in #252.

ivmai:
Thank you, Petter, for your patch anyway, it helped me to reconsider the issue.

paurkedal:
Yes, that's better, esp. the fact that the test now also runs in leak detection mode.

ivmai:
Right, even though it is somewhat useless case (I mean running most of tests in this mode).

PR #269, opened 2019-03-15
Fix compiling on OpenBSD on UltraSparc

AlaskanEmily:


ivmai:
Thank you, I will process the patch. (I will remove sparc_sunos4_mach_dep.s as well.)
AppVeyor build failure is not related to this patch.

ivmai:
Merged as c49324c

Issue #274, opened 2019-04-09
How to instruct BDWGC to scan manually-allocated stacks in coroutines


nickwanninger:
Is there a way to add a region of memory mapped with `mmap` to the garbage collector so it searches there for references?

ivmai:
Do you mean adding a data root? If yes, please look at GC_add_roots()

nickwanninger:
Hmm doesnt seem to work. I'm trying to implement a stackful coroutine system in my language which requires allocating a new c stack that I then switch in and out of (treated as python generators). Though when I use libgc I get unexpected segfaults from what I assume is incorrectly freed data. I can turn off the garbage collector and use system malloc and everything works perfectly fine.

Is there something else I need to do to specifically add a stack?

ivmai:
Similar to #150

ivmai:
At least the problem is how to determine the stack base of the current coroutine stack.
Coroutines are not supported out-of-the-box in BDWGC currently, so some tricks should be used.
Look at STACK_NOT_SCANNED.

ivmai:
Also look at GC_[un]register_my_thread.

ivmai:
I think the followings should work (STACK_NOT_SCANNED and GC_add_roots are not needed).
Each coroutine should look like:
```
void coroutine_inner(struct GC_stack_base *pbase, void *arg) {
    GC_register_my_thread(pbase); // this one records the stack base of the this coroutine
    .... // your code here
    GC_do_blocking(&yield, 0); // yield() is the function what switches to other coroutine
         // GC_do_blocking flushes all registers to stack, records current stack pointer, and instruct the GC that this thread should not be stopped during stop-the-world collection phase, and then calls yield, and it "unblocks" when yield returns execution control to this coroutine.
    .... // your code here
    GC_unregister_my_thread(); // unregisters the coroutine
}

void coroutine(void *arg) {
   GC_call_with_stack_base(coroutine_inner, arg);
}
```

nickwanninger:
This will work if I yield out of the coroutine before it is completed as well?

ivmai:
> This will work if I yield out of the coroutine before it is completed as well?

Do you mean yield and never return back (never do GC_unregister_my_thread)? In this case, it will stay registered in the collector and its stack will be scanned on every garbage collection.

nickwanninger:
[...]

PR #277, opened 2019-04-22
Added functions to get and set the stackbottom of each thread

bcardiff:
The added functions allow the implementation of lightweight threads without imposing new concepts to the user of the GC.

Hopefully, this PR is the beginning of a conversation on how to add this feature. I suspect there are things you will want to discuss before merging.

In Crystal, we are starting to use this extension as a preview feature in the path to support multi-threading.

### DIFF
--- a/include/gc.h
+++ b/include/gc.h
@@ -362,6 +362,16 @@ GC_API GC_ATTR_DEPRECATED char *GC_stackbottom;
                                 /* GC_call_with_gc_active() and         */
                                 /* GC_register_my_thread() instead.     */
 
+/* Sets the cool end of user stack in the specified thread.             */
+/* MUST be called with the GC disabled.                                 */
+GC_API void GC_CALL GC_set_stackbottom(void * thread, char * stackbottom);
+
+/* Returns the cool end of user stack of the current thread/coroutine.  */
+/* When a thread starts it will be computed upon creation.              */
+/* If GC_switch_to_coroutine() is called, the returned value will match */
+/* the specified value in that function.                                */
+GC_API char * GC_CALL GC_get_stackbottom();
+
 GC_API GC_ATTR_DEPRECATED int GC_dont_precollect;
                                 /* Do not collect as part of GC         */
                                 /* initialization.  Should be set only  */
--- a/pthread_support.c
+++ b/pthread_support.c
@@ -1369,6 +1369,44 @@ GC_INNER void GC_do_blocking_inner(ptr_t data, void * context GC_ATTR_UNUSED)
     UNLOCK();
 }
 
+/* Sets the cool end of user stack in the specified thread.             */
+/* MUST be called with the GC disabled.                                 */
+GC_API void GC_CALL GC_set_stackbottom(void * thread, char * stackbottom)
+{
+    GC_thread me;
+
+    me = GC_lookup_thread((pthread_t) thread);
+
+    if ((me -> flags & MAIN_THREAD) == 0) {
+        me -> stack_end = stackbottom;
+    } else {
+        GC_stackbottom = stackbottom;
+    }
+}
+
+/* Returns the cool end of user stack of the current thread/corutine.   */
+/* When a thread starts it will be computed upon creation.              */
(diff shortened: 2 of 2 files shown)


ivmai:
Relates to #274

ivmai:
Notes on pull requests:
* It should be based on master
* After testing it could be later back-ported to the release branches (but typically new functionality is not back-ported)

ivmai:
I was considering such solution but the issue with it that we completely forget to scan old stack. E.g. you have the only object pointer in the primordial thread's stack then you switch to a coroutine, then GC occurs which won't scan the the primordial thread's stack thus causing the live object to be collected.
How do you deal with such situations?

ivmai:
Currently I'm working on another solution described in #274. It has some drawbacks too.

bcardiff:
I'm interested in whatever built-in support bdw-gc can have that can be used in Crystal.

In Crystal:
* we already need to keep a list of all the lightweight threads in order to switch/restore their execution context. 
* The initial lightweight threads of each thread are also in that list. 
* Each element knows if it is currently bound to a thread or not.
* Before a collect, thanks to `GC_set_push_other_roots`, all the stacks of the lightweight threads that are not currently running are pushed as roots.

Although having an abstraction can be useful, at the end of the day a scheduler would need to know about the lightweight threads, so there is no real gain to have some else that the proposed feature.

A final note is that the proposed extension is only called before the collect, there is no need to let the GC know about all the context switches the runtime is doing.

ivmai:
OK. The concept looks good.
@nickwanninger, what do you think of it?

ivmai:
> Hopefully, this PR is the beginning of a conversation on how to add this feature. I suspect there are things you will want to discuss before merging.

What to improve (some items already mentioned above):
* rebase to master
* don't mention functions that are missing in GC (e.g. GC_switch_to_coroutine)
* don't duplicate comments in .h and .c (API function should be documented in .h, adding the same commit to the function definition is redundant)
* Use GC_stack_base to retrieve and update thread's stack bottom (i.e. GC_get_stackbottom should have void return type)
* The API should be cross-platform (single-threaded, pthreads, win32) - for the single-threaded case we can add a comment that 0 could be passed as a thread-id, the type for thread id is GC_SUSPEND_THREAD_ID (this is defined in javaxfc.h, so we could move GC_set_stackbottom there, the type should be DWORD if win32); to simplify cross-platform usage, I think it would be good to allow 0 as thread id to specify the current thread
* GC_set_stackbottom (i.e. user-level context switching based on GC_set_stackbottom) is not compatible GC_do_blocking and GC_call_with_gc_active currently - it is OK but there should be the relevant comment in .h and there should be assertions that GC_blocked_sp and GC_traced_stack_sect (or their multi-threaded equivalents) are null on GC_set_stackbottom entrance
* "MUST be called with the GC disabled": do you mean the caller should acquire the GC lock (by GC_call_with_alloc_lock)? We should add GC_ASSERT(I_HOLD_LOCK()) to GC_set_stackbottom in this case
* There should be some testing of get/set functionality in gctest (the easiest (minimum) thing is to get stack bottom and immediately set it to the same thread)

> A final note is that the proposed extension is only called before the collect, there is no need to let the GC know about all the context switches the runtime is doing.

Also, is my understanding correct that you listen GC events and perform GC_set_stackbottom for all threads? Is it correct that we could also call GC_set_stackbottom on each context switch (w/o listening GC events)?

ivmai:
I'm closing this PR (new version to be in another PR). But discussion is OK to continue here or in #274

nickwanninger:
Yeah, I'll look into it. I'm still having problems integrating GC into a CMake build system on MacOS though, so it may take a little bit of time

nickwanninger:
Does this PR still mark the old stack? When you switch to another stack base, is the old stack still being scanned for references?

ivmai:
No, you should do bookkeeping of the coroutines yourself and handle marking of inactive coroutines (including the initial one) using `GC_set_push_other_roots`. To fetch the bottom of the current stack you can use one of the proposed functions.

Issue #275, opened 2019-04-10
Fix cmake build on MacOS


nickwanninger:
I get this error on every single file that the CMake program attempts to compile. Is this a known problem or is my machine not configured correctly?

```
In file included from /var/folders/33/mjfjj83j3zgc_hfnrkj9wkjw0000gn/T/tmp.HMPg99oq/bdwgc/alloc.c:19:
In file included from /var/folders/33/mjfjj83j3zgc_hfnrkj9wkjw0000gn/T/tmp.HMPg99oq/bdwgc/include/private/gc_priv.h:59:
In file included from /usr/include/stdlib.h:65:
In file included from /usr/include/sys/wait.h:109:
In file included from /usr/include/sys/signal.h:150:
In file included from /usr/include/sys/_types/_sigaltstack.h:40:
/usr/include/machine/types.h:37:2: error: architecture not supported
#error architecture not supported
 ^
```

ivmai:
CMake works on Windows. Should work on other hosts but not tested for ages. The script contains some workaround for MacOS:
```
IF(APPLE)
    IF("${CMAKE_OSX_ARCHITECTURES}" STREQUAL "")
        SET(CMAKE_OSX_ARCHITECTURES "ppc;i386;x86_64" CACHE STRING "Build architectures for Mac OS X" FORCE)
    ENDIF()
ENDIF(APPLE)
```
I guess it should not be difficult to fix the build.

nickwanninger:
Hmm, I've looked into this and the only way I can get it working is by a) moving that workaround to above the `project()` call, and b) changing the architectures to only "x86_64".

```
IF(APPLE)
    IF("${CMAKE_OSX_ARCHITECTURES}" STREQUAL "")
        SET(CMAKE_OSX_ARCHITECTURES "x86_64" CACHE STRING "Build architectures for Mac OS X" FORCE)
    ENDIF()
ENDIF(APPLE)
```

Doing this allows me to fully compile using cmake on macos

ivmai:
If you know how to fix this in GC, please contribute.

vicroms:
@nickwanninger 

I've opened a PR (#281) to modify the existing CMakeLists.txt, which I believe is due for modernization. 

I'll try to add your workaround there as well.

vicroms:
Hi @nickwanninger 
It's been a while, but can you test whether the patches in PR #281 work for you?

nickwanninger:
Yes I will look into the PR really quick

[...]

PR #281, opened 2019-05-25
[CMake] Modernize CMake integration

vicroms:
Hi @ivmai 

I'm a maintainer for [**vcpkg**](https://github.com/Microsoft/vcpkg).

A vcpkg contributor (@NNemec) and I are trying to port `bdwgc` to vcpkg, see: https://github.com/microsoft/vcpkg/pull/6405.

While doing so, I've made some changes to the `CMakeLists.txt` file.

The goal is to match the output of the CMake build to the files distributed in the Debian package (which, I assume is what most downstream consumers use).

Using vcpkg`s CI, I can confirm that this results in successful builds for:

- Windows 32-bit
- Windows 64-bit
- Windows UWP (Windows 10 Store apps)
- Windows for ARM
- x64-Linux (tested with Ubuntu)

The existing `CMakeLists.txt` only produced succesful builds for Windows 32- and 64-bit.

Both the existing version and the modified version fail to build correctly on Mac OS X, but that's something that can be added in a future PR.

**Note: this only produces C++ libraries.**


ivmai:
OK, looking into it now. Seems to be good.
The thing I worry about is the style of keywords (upper/lower case)- I want to keep one for all CMake files. I don't have a preference, currently uppercase is used. If you think that lowercase is more readable or more common then it is OK to make a refactory commit.

ivmai:
lower-case seems to be more common

ivmai:
How about moving cord-specific part to Cmakefile in cord?

NNemec:
Sounds reasonable, but it would be good if you could do that as a separate effort to avoid blocking this pull-request. Generally, I would even suggest to deactivate the build of cord by default. My guess is that only a negligible fraction of gc users would be interested in a highly specialized string handling library.

ivmai:
> it would be good if you could do that as a separate effort to avoid blocking this pull-request.

OK.

> Generally, I would even suggest to deactivate the build of cord by default. My guess is that only a negligible fraction of gc users would be interested in a highly specialized string handling library.

For now, I'm not going to deactivate it, though I agree that it is of little usage. Currently it is on primarily for gc testing.

ivmai:
I will review and merge your patch in 2-3 days

vicroms:
I've solved the conflicts with `master` and added back the `add_subdirectory(tests)` which I had removed for the purposes of building the **vcpkg** port.

CMakeLists.txt
```diff
@@ -29,7 +29,14 @@ include(CTest)
 
 cmake_minimum_required(VERSION 2.6)
 
-add_definitions("-D_CRT_SECURE_NO_DEPRECATE")
+option(BDWGC_INSTALL_CORD "Install cord" ON)
+option(BDWGC_SKIP_HEADERS "Skip installing header files" OFF)
+option(BDWGC_SKIP_TESTS "Skip building tests" OFF)
+
```
vicroms: Any comments on the names of these options?
I namespaced them with a 'BDWGC_' prefix, but these prefix is not present in other options, e.g.: `enable_gc_debug`.

Also, would it be a good idea to move all of the options to the top of the file?

Right now all of the options are declared just before being used, which makes them a bit hard to find.

CMakeLists.txt
```diff
@@ -238,24 +235,59 @@ if (enable_checksums)
...
-add_library(gc-lib STATIC ${SRC})
-set_target_properties(gc-lib PROPERTIES
-                      COMPILE_DEFINITIONS GC_NOT_DLL)
-#TODO TARGET_LINK_LIBRARIES(...  ...  ${LIBS})
-
-add_library(gcmt-lib STATIC ${SRC})
-set_target_properties(gcmt-lib PROPERTIES
-                      COMPILE_DEFINITIONS GC_NOT_DLL)
+# Create target
+add_library(gc ${SRC})
```
vicroms: Usually, in modern CMake, the creation of the target would be near the top of the file.

That way, definitions can be added with `target_add_definitions()` to specific targets instead of `add_definitions()` which is global.

CMakeLists.txt
```diff
@@ -238,24 +235,59 @@ if (enable_checksums)
...
+    target_compile_definitions(gc PRIVATE GC_NOT_DLL)
+else()
+    target_compile_definitions(gc PRIVATE GC_DLL)
+endif()
 
-add_library(gcmt-dll SHARED ${SRC})
+# Install library
+install(
+    TARGETS gc
+    LIBRARY DESTINATION lib
```
vicroms: A feature to consider would be exporting the target, so that this CMakeLists generates `gc-config.cmake` and `gc-targets.cmake`, which would make linking to `gc` easier for downstream consumers. i.e.: it would allow `find_package(gc)` to work out-of-the-box.

CMakeLists.txt
```diff
@@ -238,24 +235,59 @@ if (enable_checksums)
...
+    )
+
+    install(
+        TARGETS cord 
+        DESTINATION tools/cord 
+    )
 endif()
 
-add_subdirectory(tests)
+if (NOT BDWGC_SKIP_TESTS)
```
vicroms: I added an option to control whether tests are installed here.
It is ON by default.

CMakeLists.txt
```diff
@@ -29,7 +29,14 @@ include(CTest)
 
 cmake_minimum_required(VERSION 2.6)
```
vicroms: Would you consider raising the minimum required version to 3.0?

There's much added goodness that we're missing on like `CMakePackageConfigHelpers` for easy integration in CMake projects.

ivmai:
I will reply to your questions later.

What's wrong with https://ci.appveyor.com/project/ivmai/bdwgc/build/job/j1450qvf94bqjary ?

CMakeLists.txt
ivmai: > Any comments on the names of these options?

It is better to remove BDWGC_ prefix.

> Any comments on the names of these options?

Sure. Could you please rebase your PR and do it? (It is OK to do in the same PR or open another one)

CMakeLists.txt
ivmai: OK, please add a commit or PR.

CMakeLists.txt
ivmai: OK

CMakeLists.txt
ivmai: Having an option is OK.
Test generation should be on by default. Is it OK?

CMakeLists.txt
ivmai: OK, please make a commit or PR

vicroms:
> What's wrong with https://ci.appveyor.com/project/ivmai/bdwgc/build/job/j1450qvf94bqjary ?

I'm not too familiar with how testing works, but it looks like the tests are not linking `gc` properly because they reference the old target name `gc-lib`.

I'll update the `tests/CMakeLists.txt` to use the new target name.

CMakeLists.txt
vicroms: Will do in a separate PR after this one is finished

ivmai:
I'm continuing review of your 2 patches in this PR.

CMakeLists.txt
```diff
@@ -88,10 +92,10 @@ if (CMAKE_USE_PTHREADS_INIT)
   # Common defines for most POSIX platforms.
   if (HOST MATCHES .*-.*-aix.*|.*-.*-android.*|.*-.*-cygwin.*|.*-.*-darwin.*|.*-.*-.*freebsd.*|.*-.*-haiku.*|.*-.*-gnu.*|.*-.*-hpux11.*|.*-.*-irix.*|.*-.*-.*linux.*|.*-.*-msys.*|.*-.*-nacl.*|.*-.*-netbsd.*|.*-.*-openbsd.*|.*-.*-osf.*|.*-.*-solaris.*)
     add_definitions("-DGC_THREADS -D_REENTRANT")
-    if (enable_parallel_mark)
```
ivmai: I'm going to leave lower case variable names for now (to match the configure)

ivmai:
This pull request **introduces 11 alerts** when merging af50acc98de6e7690ead61b80b7601591739d4c0 into 2b7003769e1ecc941ef80b603cba16a121f37997 - [view on LGTM.com](https://lgtm.com/projects/g/ivmai/bdwgc/rev/pr-97377805945f82c495a614f6684a4a5a808f4bc0)

**new alerts:**

* 10 for FIXME comment
* 1 for Missing header guard

---

*Comment posted by [LGTM.com](https://lgtm.com)*

CMakeLists.txt
vicroms: Got it.
Merged master into PR branch and used all lowercase names for options

vicroms:
I noticed that one of the CI tests fails with the following error:

```
libtool: link: gcc -DSTATICROOTSLIB2 -DGC_VISIBILITY_HIDDEN_SET -fvisibility=hidden -Wall -Wextra -Wpedantic -Wno-long-long -g -O2 -fno-strict-aliasing -D GCTEST_PRINT_VERBOSE -o .libs/staticrootstest.exe tests/staticrootstest-staticrootstest.o  ./.libs/libstaticrootslib_test.dll.a ./.libs/libstaticrootslib2_test.dll.a /cygdrive/c/projects/bdwgc/.libs/libgc.dll.a ./.libs/libgc.dll.a -L/nowhere -L/usr/local/lib
```

If you notice, it is liking to the old binary names `libgcc.dll.a`, while this CMakeLists should produce:
- `gc.dll`, and 
- `gc.lib` 
on Windows.

ivmai:
What does it mean "install tests"? Why do we need an option for it?

vicroms:
It is ON by default, if it is set to OFF then it skips building the tests.  It is useful in **vcpkg** so that only the library is installed without any other extra binary (like cord for example).

ivmai:
Review is still in progress.

Is it OK if I rename install_test, install_cord to build_test/cord ?

vicroms:
Yes, that's ok.

CMakeLists.txt
```diff
@@ -52,22 +56,13 @@ add_definitions("-DALL_INTERIOR_POINTERS -DNO_EXECUTE_PERMISSION")
 
 if (APPLE)
   if ("${CMAKE_OSX_ARCHITECTURES}" STREQUAL "")
-    set(CMAKE_OSX_ARCHITECTURES "ppc;i386;x86_64"
+    set(CMAKE_OSX_ARCHITECTURES "x86_64;i386"
```
ivmai: Does the presence of ppc lead to build failure on OS X?

vicroms: I'm not a Mac user myself, but this change enables Mac OS on **vcpkg**'s CI tests.

The following snippet causes a build failure when adding the `ppc` architecture:  

[`cdefs.h`](https://opensource.apple.com/source/xnu/xnu-4570.41.2/bsd/sys/cdefs.h.auto.html)
```
/*
 * Architecture validation for current SDK
 */
#if !defined(__sys_cdefs_arch_unknown__) && defined(__i386__)
#elif !defined(__sys_cdefs_arch_unknown__) && defined(__x86_64__)
#elif !defined(__sys_cdefs_arch_unknown__) && defined(__arm__)
#elif !defined(__sys_cdefs_arch_unknown__) && defined(__arm64__)
#else
#error Unsupported architecture
#endif
```

I assume that this only fails newer Mac OS versions.
I'm not sure in which version of MacOS was `ppc` support deprecated, but it is worth investigating so that we can conditionally pass the `ppc` architecture.

It is based on @nickwanniger 's fix for issue #275

ivmai: OK. It would be later to add arm and arm64 (when someone tests them)

ivmai: OK. It would be later to add arm and arm64 (when someone tests them).

nickwanninger:
I mentioned in #275, that if you run `cmake .. -DCMAKE_OSX_ARCHITECTURES=x86_64`, it seems to work with macOS, ignoring a warning:
```
CMake Warning (dev):
  Policy CMP0042 is not set: MACOSX_RPATH is enabled by default.  Run "cmake
  --help-policy CMP0042" for policy details.  Use the cmake_policy command to
  set the policy and suppress this warning.

  MACOSX_RPATH is not specified for the following targets:

   gcmt-dll

This warning is for project developers.  Use -Wno-dev to suppress it.
```

But other than that, this PR works on macOS with that flag

ivmai:
I have an issue with tests execution when shared library is built (at least on Windows): the tests are created and executed in tests folder while the library is created in base folder, so the tests fail because the library is not found.
I see 3 ways to fix it:
1. move the content of tests/CMakeLists.txt to base CMakeLists.txt
2. tell cmake somehow that the test executables should be put to the base folder
3. tell cmake/ctest somehow (working_directory?) that the tests should be launched from the base folder

Ideal way is 2 but I failed to do it; neither way 3. It could be fixed with way 1 but I'd like to keep tests/CMakeLists.txt.

Do you know how to fix it in way 2 or 3?

wangp:
Can you use `add_test(... WORKING_DIRECTORY ...)` or set the PATH environment variable?
https://cmake.org/pipermail/cmake/2015-November/061978.html
(not tested, sorry)

vicroms:
Sorry, I haven't had time to look into the issues.
Any updates so far?

ivmai:
In progress. I'll finish this week

ivmai:
Why do we need target_include_directories ?

I have pushed all the rest of the changes your requested. Please check it.

ivmai:
Related PR: https://github.com/microsoft/vcpkg/pull/20580 (update package to v8.2.0)

PR #283, opened 2019-06-06
Get rid of some non-ELF ifdefs (for now ELF platforms)

coypoop:
Hopefully improve the state of the file slightly!

OpenBSD/m68k was apparently a later change, so I left that in.


ivmai:
OK, I will accept it. But, generally, I resist to remove all old platforms support.

ivmai:
I have squashed and merged your patches but with some modifications (I left as-is some of non-ELF variants which are still officially downloadable).

Issue #296, opened 2019-10-26
Crashes with "sem_wait failed" when creating many threads on macOS


thomasthiriez:
Hi,

I have an application that uses libgc on macOS, and it randomly crashes when it creates many threads, with this message:

`sem_wait failed`

Produced on this line: [pthread_support.c#L1986](https://github.com/ivmai/bdwgc/blob/ed912875addff7e0724e5da2734b8f878c4f7148/pthread_support.c#L1986)

On macOS, the semaphore is actually a `pthread_mutex`, and it is the call to `pthread_mutex_lock` that fails with an error of `EINVAL`, there: [darwin_semaphore.h#L67](https://github.com/ivmai/bdwgc/blob/ed912875addff7e0724e5da2734b8f878c4f7148/include/private/darwin_semaphore.h#L67)

Further experimentation has revealed that two threads were using the same semaphore, and one thread has destroyed the semaphore before the other could lock it.

The semaphore is stored in a `start_info` structure. That structure is allocated by a call to `GC_INTERNAL_MALLOC` ([here](https://github.com/ivmai/bdwgc/blob/ed912875addff7e0724e5da2734b8f878c4f7148/pthread_support.c#L1906)), and later released with `GC_INTERNAL_FREE` ([here](https://github.com/ivmai/bdwgc/blob/ed912875addff7e0724e5da2734b8f878c4f7148/pthread_support.c#L1992)).

In the case where the crash occurs, two threads are created, each one of them calls `GC_INTERNAL_MALLOC` to allocate a start_info, and the `GC_INTERNAL_MALLOC` function returns the same address. I have checked that `GC_INTERNAL_FREE` has not been called between the two `GC_INTERNAL_MALLOC` calls.

So the two threads end up using the same memory area to store the semaphore, and that leads to trouble.

After studying the code, I couldn't understand why that structure was being allocated dynamically in the first place. I have modified the code to allocate it on the stack instead, and the `sem_wait failed` crash has disappeared.

Even if my solution fixes the problem, it would be nice to study `GC_INTERNAL_MALLOC` and understand why it returns twice the same address in the first place.

Regards,
Thomas Thiriez

ivmai:
> two threads are created, each one of them calls GC_INTERNAL_MALLOC to allocate a start_info

Do you mean that 2 threads have invoked GC_pthread_create (almost in parallel)?
If yes, could you please check that these threads are both registered (i.e. check GC_thread_is_registered result before calling GC_pthread_create)

[...]

PR #297, opened 2019-10-28
Allocate the start_info struct on the stack in GC_pthread_create.

thomasthiriez:
Allocating the start_info struct on the stack in the GC_pthread_create function fixes a random crash when sem_wait is called on freed memory.

Issue #296.


ivmai:
Thank you. Commited as d97a9a9b1

PR #303, opened 2020-01-25
Memory leaking check in Windows.

ssrlive:


ivmai:
Thank you. Merged as 74c0c3b

PR #306, opened 2020-02-19
Add vcpkg installation instructions

NancyLi1013:
`bdwgc` is available as a port in [vcpkg](https://github.com/microsoft/vcpkg), a C++ library manager that simplifies installation for `bdwgc` and other project dependencies. Documenting the install process here will help users get started by providing a single set of commands to build `bdwgc`, ready to be included in their projects.

We also test whether our library ports build in various configurations (dynamic, static) on various platforms (OSX, Linux, Windows: x86, x64, arm, uwp) to keep a wide coverage for users.

I'm a maintainer for vcpkg, and [here is what the port script looks like](https://github.com/microsoft/vcpkg/blob/master/ports/bdwgc/portfile.cmake). We try to keep the library maintained as close as possible to the original library.

### DIFF
--- a/README.md
+++ b/README.md
@@ -255,6 +255,15 @@ or win16 is hard.
 For machines not already mentioned, or for nonstandard compilers,
 some porting suggestions are provided [here](doc/porting.md).
 
+You can download and install bdwgc using the [vcpkg](https://github.com/Microsoft/vcpkg) dependency manager:
+
+    git clone https://github.com/Microsoft/vcpkg.git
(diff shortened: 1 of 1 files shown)


ivmai:
I've added the info to [Download](https://github.com/ivmai/bdwgc/wiki/Download) page.
I recommend to also add it to https://repology.org/project/boehm-gc/packages

NancyLi1013:
@ivmai 
Thanks for your update.

PR #307, opened 2020-03-13
Add threading libraries to bdw-gc.pc

hahnjo:
This is important when building static libraries. In that case the
application must link in all transitive dependencies. Right now using
only the flags from `pkg-config --cflags --libs .../bdw-gc.pc` results
in many undefined references to `pthread_*` functions.


ivmai:
Thank you. I will review it in 1-2 weeks. I'm busy now.

ivmai:
Merged to master as commit d2c4444fe

Issue #335, opened 2020-10-16
Undefined reference to GC_dlopen()


mahge:
Environment : WSL - Ubuntu 18.04

Steps to reproduce:
- Compile static version of _gc_ with the **CMake** configuration.
- Include gc.h in another source and compile.

Probable cause:
- Missing `gc_dlopen.c` source in the CMakeLists `SRC `variable.
  https://github.com/ivmai/bdwgc/blob/57b97be07c514fcc4b608b13768fd2bf637a5899/CMakeLists.txt#L99-L102

Issue does not happen for shared library builds because the `SRC `variable is overridden to the combined source variable _extra/gc.c_

https://github.com/ivmai/bdwgc/blob/57b97be07c514fcc4b608b13768fd2bf637a5899/CMakeLists.txt#L291

---
Related PR #336

ivmai:
Solved in #336

PR #336, opened 2020-10-16
Add missing gc_dlopen.c to CMake configuration.

mahge:
- The source file _gc_dlopen.c_ is not listed in the SRC variable of the
    CMake configuration. This leads to undefined references of GC_dlopen
    when the library is built as a STATIC library.

  - When building as shared library it does not happen because the SRC
    variable listing all source files is overridden to contain only the
    single source extra/gc.c (ehich I assume contains the missing definition

Related issue #335

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -99,7 +99,7 @@ include_directories(include)
 set(SRC alloc.c reclaim.c allchblk.c misc.c mach_dep.c os_dep.c
         mark_rts.c headers.c mark.c obj_map.c blacklst.c finalize.c
         new_hblk.c dbg_mlc.c malloc.c dyn_load.c typd_mlc.c ptr_chck.c
-        mallocx.c)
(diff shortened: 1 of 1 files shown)


mahge:
I can not tell why the CI is failing. As far as I can tell, it fails on a job that has nothing to do with CMake and/or the change in this PR. I am going to force push and try to trigger it again.

ivmai:
The change looks good, please ignore build failures. I will merge this patch after finishing with another one.

mahge:
On a related note, the file specific.c is also missing from the CMakeLists file. I did not put it in this PR because I did not find a symptom for it. 

It should probably be added as well.

ivmai:
Thank you. I've merged you patch with a change - gc_dlopen.c is needed only if building with pthreads.

Regarding specific.c - yes, good catch, I've added it too in another patch - commit 6668de8

PR #338, opened 2020-11-20
Add support of FreeBSD/RISC-V

yasuhirokimura:
I'm maintainer of FreeBSD BDWGC port. Recently I received bug report from a user that the port can't be built on FreeBSD/RISC-V environment. I checked source code and found that currently Linux is only supported OS of RISC-V architecture. So I replied that currently FreeBSD/RISC-V isn't supported but if someone provides patch then I'll accept it. Then another user provided it and he said that it passes `make check` test. Since I don't have way to access FreeBSD/RISC-V environment, I can't test it by myself. But anyway I decided to create pull request as feedback to upstream.

### DIFF
--- a/include/private/gcconfig.h
+++ b/include/private/gcconfig.h
@@ -663,7 +663,7 @@ EXTERN_C_BEGIN
 #   endif
 #   define mach_type_known
 # endif
-# if defined(__riscv) && defined(LINUX)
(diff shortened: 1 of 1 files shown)


ivmai:
Thank you. I merged it with some reformatting. I have no ability to test it neither.

Issue #198, opened 2018-01-24
compilation on OpenBSD 6.2 without thread support


wangp:
The GC does not compile on OpenBSD 6.2 with thread support disabled:

```
extra/../os_dep.c:1245:16: error: use of undeclared identifier 'USRSTACK'
      result = STACKBOTTOM;
               ^
./include/private/gcconfig.h:2406:38: note: expanded from macro 'STACKBOTTOM'
#         define STACKBOTTOM ((ptr_t)USRSTACK)
                                     ^
```

`USRSTACK` is defined in `<machine/vmparam.h>` but that is protected by `_KERNEL` in `<uvm/uvm_param.h>`:

```
#ifdef _KERNEL

#include <machine/vmparam.h>
```

`make check` passes using `HEURISTIC2` instead of `STACKBOTTOM` though.

ivmai:
I will guard each STACKBOTTOM definition with "ifdef USRSTACK".

References:
* https://github.com/lattera/openbsd/blob/master/sys/arch/hppa64/include/vmparam.h

ivmai:
Please test the above patch.

wangp:
Thanks. We still need to define HEURISTIC2. HEURISTIC1 does not work. I only tested on x86_64.

ivmai:
Fixed.

PR #347, opened 2021-02-10
Update OpenBSD Support

bsdkurt:
This pull request updates OpenBSD support and removes some outdated workarounds. This is based on the recent OpenBSD update in the ports tree (openbsd/ports@249e6c0) with the following differences:

I've completely removed support for OpenBSD uthreads. OpenBSD user-land threading model was replaced by kernel supported threads (rthreads) about 9 years ago so it seems like keeping this dead code around is not useful anymore.

When threading is disabled by configure, USERSTACK is not usable on OpenBSD due to the random sized PROT_NONE stack gap. So I removed its use and will always use HEURISTIC2. when threading is disabled on OpenBSD.

### DIFF
--- a/dyn_load.c
+++ b/dyn_load.c
@@ -81,13 +81,6 @@ STATIC GC_has_static_roots_func GC_has_static_roots = 0;
 #   define ELFSIZE ARCH_ELFSIZE
 #endif
 
-#if defined(OPENBSD)
-# include <sys/param.h>
-# if (OpenBSD >= 200519) && !defined(HAVE_DL_ITERATE_PHDR)
-#   define HAVE_DL_ITERATE_PHDR
-# endif
-#endif /* OPENBSD */
-
 #if defined(SCO_ELF) || defined(DGUX) || defined(HURD) || defined(NACL) \
     || (defined(__ELF__) && (defined(LINUX) || defined(FREEBSD) \
                              || defined(NETBSD) || defined(OPENBSD)))
@@ -149,8 +142,10 @@ STATIC GC_has_static_roots_func GC_has_static_roots = 0;
 #    elif defined(NETBSD) || defined(OPENBSD)
 #      if ELFSIZE == 32
 #        define ElfW(type) Elf32_##type
-#      else
+#      elif ELFSIZE == 64
 #        define ElfW(type) Elf64_##type
+#      else
+#        error Missing required ELFSIZE define
 #      endif
 #    else
 #      if !defined(ELF_CLASS) || ELF_CLASS == ELFCLASS32
--- a/include/gc_config_macros.h
+++ b/include/gc_config_macros.h
@@ -84,13 +84,13 @@
 #elif defined(GC_THREADS)
 # if defined(__linux__)
 #   define GC_LINUX_THREADS
+# elif defined(__OpenBSD__)
+#   define GC_OPENBSD_THREADS
 # elif defined(_PA_RISC1_1) || defined(_PA_RISC2_0) || defined(hppa) \
        || defined(__HPPA) || (defined(__ia64) && defined(_HPUX_SOURCE))
 #   define GC_HPUX_THREADS
 # elif defined(__HAIKU__)
 #   define GC_HAIKU_THREADS
-# elif defined(__OpenBSD__)
-#   define GC_OPENBSD_THREADS
 # elif defined(__DragonFly__) || defined(__FreeBSD_kernel__) \
        || (defined(__FreeBSD__) && !defined(SN_TARGET_ORBIS))
 #   define GC_FREEBSD_THREADS
--- a/include/private/gc_priv.h
+++ b/include/private/gc_priv.h
@@ -2795,9 +2795,7 @@ GC_INNER void *GC_store_debug_info_inner(void *p, word sz, const char *str,
 #     define SIG_SUSPEND SIGPWR
 #   endif
 # elif defined(GC_OPENBSD_THREADS)
-#   ifndef GC_OPENBSD_UTHREADS
-#     define SIG_SUSPEND SIGXFSZ
-#   endif
+#   define SIG_SUSPEND SIGXFSZ
 # elif defined(_SIGRTMIN) && !defined(CPPCHECK)
 #   define SIG_SUSPEND _SIGRTMIN + 6
 # else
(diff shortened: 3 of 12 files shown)


ivmai:
I'm back to reviewing the patches in this PR - 3 of 7 are merged.

dyn_load.c
```diff
@@ -81,13 +81,6 @@ STATIC GC_has_static_roots_func GC_has_static_roots = 0;
 #   define ELFSIZE ARCH_ELFSIZE
 #endif
 
-#if defined(OPENBSD)
```
ivmai, line 84: Why do you remove this block? HAVE_DL_ITERATE_PHDR should not be defined for latest OpenBSD? Or it is always defined by configure?

ivmai:
Commit 9e45f5ad relates to issue #198. @wangp, do you agree with the commit?

bsdkurt:
> I'm back to reviewing the patches in this PR - 3 of 7 are merged.

Great. Thank you.

dyn_load.c
bsdkurt: It is always defined by configure now.

ivmai: Got it. I prefer to keep this code for now. Because 4 build systems are supported (for OpenBSD), but HAVE_DL_ITERATE_PHDR is provided only in the 1st two of them:
* configure
* CMake
* Makefile.direct
* none (e.g. gcc -I include extra/gc.c)

.

bsdkurt: Okay. This was more of a cleanup of what I thought was unneeded code. Keeping it is fine as well.

bsdkurt:
> Commit [9e45f5a](https://github.com/ivmai/bdwgc/commit/9e45f5ad77cc338960a18c9414e328ac7bef1a4a) relates to issue #198. @wangp, do you agree with the commit?

Some additional context for this change; I noticed USRSTACK was visible to user-land on HPPA (if I recall correctly). So I tested the build without threading and it failed the tests. The reason is that on all architectures that OpenBSD supports, the start of stack for the executable is randomly placed some distance away from USRSTACK with PROT_NONE pages between the start and USRSTACK. So when threading is disabled HEURISTIC2 is needed to find the actual start of the stack even if USRSTACK is visible to user-land. After making this change HPPA without threads passed the tests.

bsdkurt:
It looks like all that is outstanding in this pull request is the removal of user-land threading support in commit a4bcba1

ivmai:
> It looks like all that is outstanding in this pull request is the removal of user-land threading support in commit [a4bcba1](https://github.com/ivmai/bdwgc/commit/a4bcba149199ab03feed6f84ecc25b2a7f94f88b)

I'm looking at it now. Besides removal of userland threads support you've also replaced GC_find_limit_openbsd to GC_find_limit_with_bound in the same commit. It seems to be OK to do this replacement, I have no ability to test though, but we need to update GC_set_and_save_fault_handler too:
|| defined(NETBSD)
->
|| defined(NETBSD) || defined(OPENBSD)

And probably use SA_NODEFER.

What do you think? (no need to update the PR, I need you opinion and testing)

bsdkurt:
> > It looks like all that is outstanding in this pull request is the removal of user-land threading support in commit [a4bcba1](https://github.com/ivmai/bdwgc/commit/a4bcba149199ab03feed6f84ecc25b2a7f94f88b)
> 
> I'm looking at it now. Besides removal of userland threads support you've also replaced GC_find_limit_openbsd to GC_find_limit_with_bound in the same commit. It seems to be OK to do this replacement, I have no ability to test though, but we need to update GC_set_and_save_fault_handler too:
> || defined(NETBSD)
> ->
> || defined(NETBSD) || defined(OPENBSD)
> 
> And probably use SA_NODEFER.
> 
> What do you think? (no need to update the PR, I need you opinion and testing)

Thank you for the thorough review. The GC_find_limit_openbsd work-around was explicitly written for uthreads, however it wasn't properly ifdef'ed to only apply to uthreads which was an oversight. In the current openbsd port of 8.0.4, I corrected the ifdef's so that GC_find_limit_openbsd is only used for the GC_OPENBSD_UTHREADS case (dead code now):

https://github.com/openbsd/ports/blob/master/devel/boehm-gc/patches/patch-os_dep_c#L50

This comment describes why it was needed for uthreads:
```
  /* Don't use GC_find_limit() because siglongjmp() outside of the      */
  /* signal handler by-passes our userland pthreads lib, leaving        */
  /* SIGSEGV and SIGPROF masked.  Instead, use this custom one that     */
  /* works-around the issues.                                           */
```

The 8.0.4 port on openbsd is working well using GC_find_limit_with_bound on all our architectures.

I think updating GC_set_and_save_fault_handler is needed to pick up SA_RESTART but is unrelated to the uthreads removal and probably should be a separate commit. I'm looking into this a bit to determine if:

- SA_NODEFER is needed or not (I suspect not)
- If a SIGBUS handler is needed or not (I suspect not but need to test more)

ivmai:
> If a SIGBUS handler is needed or not (I suspect not but need to test more)

SIGBUS was not used in GC_find_limit_openbsd.

ivmai:
>  The GC_find_limit_openbsd work-around was explicitly written for uthreads, however it wasn't properly ifdef'ed to only apply to uthreads which was an oversight.

Okay, I switched to GC_find_limit_with_bound (using sigaction but w/o SA_NODEFER), see commit 9775d93.

> I'm looking into this a bit to determine if: SA_NODEFER is needed or not (I suspect not)

Okay, if it is needed please open new PR.
Thank you

ivmai:
The request to remove uthreads support is moved to issue #358. It should not a be a problem to remove the code which is unlikely to be used but I prefer to be conservative regarding it. E.g. for the case we need similar algorithm for some other target.

bsdkurt:
@ivmai Thank you again for the review and merging of the updated support. I will test the master branch on OpenBSD i386/amd64/aarch64/sparc64 next week. I have hppa/macppc as well but they are powered off at a different location and will take longer to test.

bsdkurt:
FYI: i386/amd64/aarch64/sparc64 tested and working well on OpenBSD.

ivmai:
> FYI: i386/amd64/aarch64/sparc64 tested and working well on OpenBSD.

Good. Thank you!

ivmai:
> @bsdkurt
[Remove OpenBSD uthreads support. Userland threads were replaced by](https://github.com/ivmai/bdwgc/pull/347/commits/a4bcba149199ab03feed6f84ecc25b2a7f94f88b) 
[a4bcba1](https://github.com/ivmai/bdwgc/pull/347/commits/a4bcba149199ab03feed6f84ecc25b2a7f94f88b)

Okay, after repetitive requests to remove uthreads support, I've merged the commit (a similar one as your one but for the fresh master) - commit 2c1ae2ea4

bsdkurt:
@ivmai Thank you

PR #352, opened 2021-04-17
Mac little build update GC_FindTopOfStack, changes aimed for Mac M1/iOs.

devnexen:
The frame pointer register is fp is that platform.

### DIFF
--- a/darwin_stop_world.c
+++ b/darwin_stop_world.c
@@ -70,7 +70,7 @@ GC_INNER ptr_t GC_FindTopOfStack(unsigned long stack_start)
         frame = (StackFrame *)sp_reg;
 #   elif defined(AARCH64)
         volatile ptr_t sp_reg;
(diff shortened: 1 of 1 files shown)


darwin_stop_world.c
```diff
@@ -70,7 +70,7 @@ GC_INNER ptr_t GC_FindTopOfStack(unsigned long stack_start)
         frame = (StackFrame *)sp_reg;
 #   elif defined(AARCH64)
         volatile ptr_t sp_reg;
-        __asm__ __volatile__ ("mov %0, x29\n" : "=r" (sp_reg));
+        __asm__ __volatile__ ("mov %0, fp\n" : "=r" (sp_reg));
```
ivmai, line 73: Why do we need this change? x29 is frame pointer register (see https://developer.apple.com/documentation/xcode/writing-arm64-code-for-apple-platforms)

PR #354, opened 2021-05-12
Emscripten support with gctest mostly passing

pclouds:
This adds a bit more in os_dep.c in order to scan the stack and
registers, which is possible if we enable ASYNCIFY (note, this adds
quite some overhead). gctest almost passes with this, except the tests
that involve a_get()

The build system (at least autotools-based) is updated to detect and
configure emscripten correctly. No thread support for now. No idea how
to stop other threads (perhaps we need support from JS side?)

To build

    # source EMSDK first
    emconfigure ./configure
    emmake make gctest.html
    # point your browser at .libs/gctest.html

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -91,6 +91,10 @@ endif
 endif
 endif
 
+if EMSCRIPTEN
+libgc_la_SOURCES += emscripten.c
+endif
+
 if THREAD_LOCAL_ALLOC
 libgc_la_SOURCES += thread_local_alloc.c
 endif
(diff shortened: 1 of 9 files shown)


ivmai:
Thank you. I've merged it with some modifications. Please test.

pclouds:
Thanks. Note that it's still not fully working though. Those failed tests caught something. I tried to use this with Gauche Scheme and it crashed unless I disabled the collection phase.

ivmai:
It's OK, more patches are welcomed!

pclouds:
That's the problem. I'm at the point where I don't know how to even debug this (and I don't know bdwgc well to begin with). Perhaps the problem is with this asyncify feature of wasm...

PR #355, opened 2021-05-27
Include <alloca.h> when using alloca on AIX

Helflym:
Using "#pragma alloca" only works with XLC, in order to enable alloca
calls. Including <alloca.h> works for both GCC and XLC

* dyn_load.c: Replace "#pragma alloca" by "#include <alloca.h>"


ivmai:
Thank you. I merged it.

PR #389, opened 2021-11-05
Distribute missing headers in single-obj-compilation

### LINKED ISSUE #388: Missing headers like gc_pthread_redirects.h if installing with --enable-single-obj-compilation
Some header files, like `include/gc_pthread_redirects.h` are missing if installing with `--enable-single-obj-compilation` (the default) even though options like `--enable-threads=posix` were specified.

The problem is that 06e4efff9949a670d41e4455ca1d4eee3d9ba58c put the specification of header inclusion inside the `!SINGLE_GC_OBJ` block in `Makefile.am`, which means that these headers won't ever be included in the installed package if using single object compilation.

jakub-w:
Makefile.am [PTHREADS] (pkginclude_HEADERS): Move outside the
SINGLE_GC_OBJ conditional.
Makefile.am [ENABLE_GCJ_SUPPORT] (pkginclude_HEADERS): Move outside
the SINGLE_GC_OBJ conditional.
Makefile.am [ENABLE_DISCLAIM] (pkginclude_HEADERS): Move outside the
SINGLE_GC_OBJ conditional.

Fixes #388

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -83,7 +83,6 @@ else
 if PTHREADS
 # Not Cygwin or MinGW.
 libgc_la_SOURCES += pthread_start.c pthread_support.c
-pkginclude_HEADERS += include/gc_pthread_redirects.h
 if DARWIN_THREADS
 libgc_la_SOURCES += darwin_stop_world.c
 else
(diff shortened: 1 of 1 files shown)


ivmai:
Thank you. Merged (with some changes in commit messages).

PR #444, opened 2022-05-02
Replace SSH cloning with HTTPS cloning in README.md

Zopolis4:
Currently the readme gives a SSH url, which is unnecessary and will result in the following error on an unconfigured machine:
```
fatal: remote error:
  The unauthenticated git protocol on port 9418 is no longer supported.
Please see https://github.blog/2021-09-01-improving-git-protocol-security-github/ for more information.
```


ivmai:
Thank you. Committed as 3667f0c59

PR #446, opened 2022-05-04
GC_remove_roots now exists on all platforms (as a no-operation on some platforms).

DoctorNoobingstoneIPresume:
This allows applications to call `GC_remove_roots` without the specific test (and without keeping sync with the specific test).

Before:

```
#if ! defined (MSWIN32) && ! defined (MSWINCE) && ! defined (CYGWIN32)
    GC_remove_roots (my_arg_b, my_arg_e);
#endif
```

Now:

```
    GC_remove_roots (my_arg_b, my_arg_e);
```

The effect is the same (no-operation on Windows and Cygwin), but it is easier to write application code.

### DIFF
--- a/mark_rts.c
+++ b/mark_rts.c
@@ -373,6 +373,9 @@ STATIC void GC_remove_tmp_roots(void)
     if (rebuild)
         GC_rebuild_root_index();
   }
+#else
+  GC_API void GC_CALL GC_remove_roots(void *, void *)
+  {}
 #endif /* !defined(MSWIN32) && !defined(MSWINCE) && !defined(CYGWIN32) */
 
 #ifdef USE_PROC_FOR_LIBRARIES


ivmai:
PS. Of course, it would be better to implement the function for Windows too, but it's not trivial.

DoctorNoobingstoneIPresume:
> Could you please explain more about practical need of the proposed change? What's the purpose of using GC_remove_roots on platforms where the root regions are not removed. To my understanding the if the client request removal of the region, it expects the regions would not be accessed (by bdwgc) after GC_remove_roots (e.g. the client could unmap the memory after the call).

Thank you for quick reply and kind information.

I have been trying to build https://github.com/p4lang/p4c (https://github.com/ivmai/bdwgc/wiki/Known-clients) on Cygwin.

Code in https://github.com/p4lang/p4c/blob/main/ir/visitor.cpp unconditionally calls `GC_remove_roots`, which is unavailable for Cygwin (and, I understand, for Windows).

This pull request has been a first superficial attempt to work-around the missing implementation.

I have just devised a second attempt: https://github.com/ivmai/bdwgc/pull/447. (-:

DoctorNoobingstoneIPresume:
> PS. Of course, it would be better to implement the function for Windows too, but it's not trivial.

Do you happen to have some pointers in this direction, please ? Perhaps something already written ? (Not that I am going to be capable of doing it -- or bypass my own laziness !)

ivmai:
Okay, thank you, I see the issue now, let me check the difficulty of supporting GC_remote_roots on Win first.

ivmai:
I agree to support GC_remove_roots() on all platforms, see commit 000f4575.
Note that I've also refined the documentation for GC_add_roots and GC_remove_roots.

DoctorNoobingstoneIPresume:
Thank you, @ivmai, with all of my heart ! >:D<

ivmai:
> support GC_remove_roots() on all platforms, see commit 000f4575c

It is also cherry-picked to release-8_2 branch (as commit aeb16bba8)

DoctorNoobingstoneIPresume:
In the meantime, I have grown to suspect that my initial plan to elide the call to `GC_remove_roots` on some platforms was probably not safe. (-: => Thank you so much more for supporting it for all platforms now !

ivmai:
Okay, thank you anyway for pointing to the issue with GC_remove_roots.
You are welcome!

PR #474, opened 2022-09-13
Fix configure with Clang 15 (implicit function declarations)

thesamesam:
Clang 15 makes implicit function declarations an error by default which leads to configure falsely thinking `pthread_setname_np` is not present:
```
checking for pthread_setname_np... no
```

This fixes that issue and the following errors:
```
error: call to undeclared function 'pthread_setname_np'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
error: call to undeclared function 'pthread_setname_np'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
```

Signed-off-by: Sam James <sam@gentoo.org>

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -834,12 +834,16 @@ AS_IF([test "$THREADS" = posix],
   [AC_MSG_CHECKING(for pthread_setname_np)
    old_CFLAGS="$CFLAGS"
    CFLAGS="$CFLAGS $CFLAGS_EXTRA -Werror"
-   AC_TRY_COMPILE([#include <pthread.h>],
+   AC_TRY_COMPILE([
+#define _GNU_SOURCE 1
+#include <pthread.h>],
                   [pthread_setname_np("thread-name")],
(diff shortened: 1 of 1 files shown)


thesamesam:
I can make another PR for master (the change is needed there) or is it easier for you to cherry pick? Whatever works best. Thanks!

ivmai:
Thank you (it would not be easy for me to discover the issue because it does not lead to configure fail). I will cherry-pick it to master myself, and some time later I will apply it to release-8_2 branch along with some other fixes.
I think I will slightly modify your patch to guard define with an ifndef.

ivmai:
Submitted as [4489757](https://github.com/ivmai/bdwgc/commit/4489757f1cc1fe4aa98d4c1ebbe0789b2b308e11). Please test it.
I'll cherry-pick it to release-8_2 later. Closing the PR.

thesamesam:
Thank you, I'll test it shortly!

PR #483, opened 2022-10-19
Allow passing in CFLAGS(_EXTRA)

plajjan:
By using ?= we only set the CFLAGS variable in case it doesn't already have a value. This means it is now possible for someone to control the way bdwgc is compiled by feeding in the CFLAGS variable whereas previously one had to modify the Makefile.direct file. Further, since we consider the CFLAGS a reasonable default, the primary means of injecting control flags is by setting CFLAGS_EXTRA, where we now also allow passing in a value and the stuff in Makefile.direct is appended to that input.

This is very useful for using bdwgc from another repo, like you can just mirror bdwgc and then compile by feeding in all the compilation options as environment variables rather than having to first write them to the Makefile.direct file, which can be rather tricky to do programmatically. The normal mode of working is probably to CFLAGS_EXTRA, since that gets merged with the CFLAGS in Makefile.direct, but it also allows to directly set CFLAGS, thus overriding the default value of CFLAGS for total control.

### DIFF
--- a/Makefile.direct
+++ b/Makefile.direct
@@ -45,9 +45,9 @@ VPATH= $(srcdir)
 # Path to atomic_ops source.
 AO_SRC_DIR=$(srcdir)/libatomic_ops
 
-CFLAGS_EXTRA=
+CFLAGS_EXTRA+=
 # We need CFLAGS_FOR_PIC because we might be building a shared library.
-CFLAGS= -O -I$(srcdir)/include -I$(AO_SRC_DIR)/src \
+CFLAGS?= -O -I$(srcdir)/include -I$(AO_SRC_DIR)/src \
   -DALL_INTERIOR_POINTERS -DENABLE_DISCLAIM -DGC_ATOMIC_UNCOLLECTABLE \
   -DGC_GCJ_SUPPORT -DJAVA_FINALIZATION -DNO_EXECUTE_PERMISSION \
   -DUSE_MMAP -DUSE_MUNMAP $(CFLAGS_FOR_PIC) $(CFLAGS_EXTRA)


ivmai:
I've modified your commit a bit using ?= for CFLAGS_EXTRA, as well as for CXXFLAGS and ABI_FLAGS. Will push it today.

ivmai:
Merged as f7cfa9d

plajjan:
Cool, thanks! =)

ivmai:
FreeBSD build failed: https://app.travis-ci.com/github/ivmai/bdwgc/jobs/586297033

plajjan:
Without looking at any core or errors, I'm guessing it is setting CFLAGS which are now used?

Guess test environment need to set whatever should be tested...

ivmai:
Strange but CFGLAGS are empty when passed to the compiler (cc).

ivmai:
Last good build (bbe9092) before the change: https://app.travis-ci.com/github/ivmai/bdwgc/jobs/584941687

ivmai:
Before (success):
```
cc  -O -I./include -I./libatomic_ops/src  -DALL_INTERIOR_POINTERS -DENABLE_DISCLAIM -DGC_ATOMIC_UNCOLLECTABLE  -DGC_GCJ_SUPPORT -DJAVA_FINALIZATION -DNO_EXECUTE_PERMISSION  -DUSE_MMAP -DUSE_MUNMAP   -c -I. ./cord/cordbscs.c
```
After (fail):
```
cc  -O2 -pipe -c -I. ./cord/cordbscs.c
```

ivmai:
It looks like as if CFLAGS="-O2 -pipe" is passed to make, but I don't understand why.

ivmai:
It looks like that the system defines default CFLAGS (probably in "/etc/portage/make.conf" as said in https://wiki.gentoo.org/wiki/GCC_optimization#What_are_CFLAGS_and_CXXFLAGS.3F).
Previously C[XX]FLAGS were overwritten in Makefile.direct.

ivmai:
I'm working on the fix, along with some more refactoring

ivmai:
Should be fixed now. The required -I and other options are appended to CFLAGS,

From now on, if you need some custom -D passed to Makefile.direct, it is better to set CFLAGS_DEFAULT_MACROS.

PR #506, opened 2022-11-22
Make Emscripten ASYNCIFY feature optional to target.

juj:
The commit https://github.com/ivmai/bdwgc/commit/1431bda1a87d66d669faab00aefe34349c54a56a# unconditionally added a requirement that Emscripten's -sASYNCIFY feature would be used.

-sASYNCIFY is a relatively rarely used feature of Emscripten, most developers do not use it, and it does not scale well to moderate-to-large codebases. It incurs a heavy impact to code size and performance, and carries other correctness problems that developers must then adhere to regarding event loop message ordering.

The commit above regressed the non-ASYNCIFY use of bdwgc, which was to limit running GC between browser event loop ticks when there is no native code on the stack. (the commit simply removed the comment that explained this use case: https://github.com/ivmai/bdwgc/commit/1431bda1a87d66d669faab00aefe34349c54a56a#diff-e328c3468f5194bb02188ac05d707e86746440c5716394e4cb9054cd91d0865d and switched to assuming that everyone would use -sASYNCIFY).

To resolve both use cases, I am proposing to add a build option `EMSCRIPTEN_ASYNCIFY` to choose whether this feature is being targeted or not. This restores the previous use case for bdwgc.

Before this change (but with the mmap fix of #505), gctest.js and middletest.js would fail with error messages

```
C:\code\bdwgc\emcc>emcmake cmake .. -Denable_threads=OFF -Denable_tests=ON -G Ninja
...
C:\code\bdwgc\emcc>ninja
...
C:\code\bdwgc\emcc>node gctest.js

C:\code\bdwgc\emcc\gctest.js:144
      throw ex;
      ^
Please compile your program with async support in order to use asynchronous operations like emscripten_scan_registers
(Use `node --trace-uncaught ...` to show where the exception was thrown)

C:\code\bdwgc\emcc>node middletest.js

C:\code\bdwgc\emcc\middletest.js:144
      throw ex;
      ^
Please compile your program with async support in order to use asynchronous operations like emscripten_scan_registers
(Use `node --trace-uncaught ...` to show where the exception was thrown)
```
since the tests were being compiled without ASYNCIFY flags enabled.

After this change, building without `-Demscripten_asyncify` results in:

```
C:\code\bdwgc\emcc>emcmake cmake .. -Denable_threads=OFF -Dbuild_tests=ON -Demscripten_asyncify=OFF -G Ninja
...
C:\code\bdwgc\emcc>ninja
...
C:\code\bdwgc\emcc>node middletest.js
[...]

PR #507, opened 2022-11-22
Fix GC_valloc() to work on Emscripten/WebAssembly.

juj:
[Physical page size in WebAssembly standard is 64KB](https://developer.mozilla.org/en-US/docs/WebAssembly/JavaScript_interface/Memory/Memory#syntax).

Bdwgc requires that the allocation alignment of a memory allocation request in `GC_memalign()` must be <= HBLKSIZE:

https://github.com/ivmai/bdwgc/blob/400053e428fcd81cf0c867d49223d74de27ccfaa/mallocx.c#L508

This check results in a call to `GC_valloc()` in gctest.js to abort when built to Wasm.

According to documentation, HBLKSIZE is capped to max out at 16KB:

https://github.com/ivmai/bdwgc/blob/400053e428fcd81cf0c867d49223d74de27ccfaa/doc/README.macros#L301-L302

It seems that simplest is to restrict the page size for WebAssembly builds to a fictional 4KB. This fixes the function `GC_valloc()` (which attempts to allocate at physical page alignment) to work in WebAssembly/Emscripten.

Combining this PR with #505 and #506 fixes gctest.js to pass when ASYNCIFY is enabled.

This PR adds a new minimal test for GC_valloc(), which eases observing this kind of issue for Emscripten. This may become useful in the future when WebAssembly standard does add virtual memory support ( https://github.com/dtig/memory-control ) .

### DIFF
--- a/include/private/gcconfig.h
+++ b/include/private/gcconfig.h
@@ -1288,6 +1288,14 @@ EXTERN_C_BEGIN
 #     define OS_TYPE "EMSCRIPTEN"
 #     define DATASTART (ptr_t)ALIGNMENT
 #     define DATAEND (ptr_t)ALIGNMENT
+      /* The real page size in WebAssembly is 64KB. However bdwgc does  */
+      /* not support a page size that high, the largest supported page  */
+      /* size is equal to HBLKSIZE (max <= 16KB). Since there is        */
+      /* currently no real virtual memory/page support in Wasm          */
+      /* (see https://github.com/WebAssembly/design/issues/1397 ), the  */
+      /* actual page size does not matter much, so override a fake 4KB  */
+      /* page size for Wasm.                                            */
+#     define GETPAGESIZE() 4096
 #     define USE_MMAP_ANON      /* avoid /dev/zero, not supported */
 #     undef USE_MUNMAP /* mmap(PROT_NONE) is unsupported, mprotect is no-op */
 #   endif


CMakeLists.txt
```diff
@@ -755,6 +755,10 @@ if (build_tests)
   target_link_libraries(middletest PRIVATE gc)
   add_test(NAME middletest COMMAND middletest)
 
+  add_executable(GC_valloctest tests/GC_valloc.c ${NODIST_SRC})
```
ivmai, line 758: Let's drop adding an extra test, GC_valloc() is already tested in gctest. (I know the bdwgc test system is not great, but I resist to redesign it for now.)

juj: ok, updated.

ivmai:
> Combining this PR with https://github.com/ivmai/bdwgc/pull/505 and https://github.com/ivmai/bdwgc/pull/506 fixes gctest.js to pass when ASYNCIFY is enabled.

If possible, please remove "ifndef EMSCRIPTEN" in gctest.c

juj:
> If possible, please remove "ifndef EMSCRIPTEN" in gctest.c

I'll look to tend to those in a followup PR to see if they are no longer necessary, if that is ok?

ivmai:
> > If possible, please remove "ifndef EMSCRIPTEN" in gctest.c
> 
> I'll look to tend to those in a followup PR to see if they are no longer necessary, if that is ok?

Sure

ivmai:
I've fixed it in a slightly different way - by commits fda6160da and 616f2a790. Please test it.

ivmai:
I've enabled large alignments (#510) and this this patch is now reverted (by ba13736).
Please test it.

PR #512, opened 2022-11-24
Remove GC_OPENBSD_UTHREADS support

blackgnezdo:
This will never be relevant now that 2012 is well behind the 1 year support horizon for OpenBSD.

### DIFF
--- a/include/private/gc_priv.h
+++ b/include/private/gc_priv.h
@@ -3023,9 +3023,7 @@ GC_INNER void *GC_store_debug_info_inner(void *p, word sz, const char *str,
 #   define SIG_SUSPEND SIGUSR1
         /* SIGTSTP and SIGCONT could be used alternatively on FreeBSD.  */
 # elif defined(GC_OPENBSD_THREADS) && !defined(GC_USESIGRT_SIGNALS)
-#   ifndef GC_OPENBSD_UTHREADS
 #     define SIG_SUSPEND SIGXFSZ
-#   endif
 # elif defined(_SIGRTMIN) && !defined(CPPCHECK)
 #   define SIG_SUSPEND _SIGRTMIN + 6
 # else
--- a/include/private/gcconfig.h
+++ b/include/private/gcconfig.h
@@ -2512,11 +2512,6 @@ EXTERN_C_BEGIN
   EXTERN_C_END
 # include <sys/param.h>
   EXTERN_C_BEGIN
-  /* Prior to 5.2 release, OpenBSD had user threads and required        */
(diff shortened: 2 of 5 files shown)


blackgnezdo:
I tested this on OpenBSD-7.2-current on amd64 by running these commands:
```
AUTOMAKE_VERSION=1.16 AUTOCONF_VERSION=2.71 ./autogen.sh && \
./configure --with-libatomic-ops=no  --enable-cplusplus \
  --enable-threads=pthreads  --enable-mmap  --enable-static=yes && \
gmake
gmake check
```

ivmai:
Thank you for the PR.
Is there any chance OpenBSD to return support of  that ancient functionality? Removing of this code should not be difficult. But signal-based approach of stopping threads is bad one compared to that one in ancient OpenBSD.

ivmai:
Similar PR to remove the support of ancient OpenBSD - https://github.com/ivmai/bdwgc/pull/347/commits/a4bcba149199ab03feed6f84ecc25b2a7f94f88b

ivmai:
I'd prefer to keep this code (under ifdef) for now just in case it would be needed in the future for some other OS.

blackgnezdo:
No, I can't see how uthreads could return in the future. And even if that happens, the extra work of reverting this commit will be lower than the drag in the meantime.

blackgnezdo:
> I'd prefer to keep this code (under ifdef) for now just in case it would be needed in the future for some other OS.

This code is never compiled, so the difference is it rots while confusing people. I started on this patch because I was confused by the usage of `__syscall`.

ivmai:
Okay, I got your point. Let me think more about.

ivmai:
Is my understanding correct that stopping threads with pthread_suspend_np is no longer supported by OpenBSD (i.e. passing -DGC_OPENBSD_THREADS to CFLAGS results in a compilation or runtime failure?

blackgnezdo:
`pthread_suspend_np` is [nowhere in the OpenBSD source tree](https://github.com/openbsd/src/search?q=pthread_suspend_np&type=code). Though I was unable to cause any failures by plugging  -DGC_OPENBSD_THREADS into CFLAGS.

ivmai:
> `pthread_suspend_np` is [nowhere in the OpenBSD source tree](https://github.com/openbsd/src/search?q=pthread_suspend_np&type=code). 

Got it.

ivmai:
I've merged your patch after some reformatting.

blackgnezdo:
Great, thanks. Appreciate the shout-out among authors.

PR #517, opened 2022-12-09
use overview.md as the "default" for all the doc files.

rdp:


ivmai:
I don't like the idea of renaming overview.md to readme.md because we already have many readme (including README.md). Probably renaming to some other unique name is OK, but not sure if it would make the documentation more evident to the beginners.
I've accepted the remaining of the PR, with some modifications - committed as [8fd5c0d](https://github.com/ivmai/bdwgc/commit/8fd5c0d9787c1bc01d45964014e866d3a3f8ddf5).
Thank you

rdp:
Yeah the initial confusion I was trying to avoid is that here: https://github.com/ivmai/bdwgc/tree/master/doc github by default displays "readme.autoconf" which is fairly confusing...maybe I could add a README.md there that says "see overview.md for an explanation of the documentation files here"?

ivmai:
> Yeah the initial confusion I was trying to avoid is that here: https://github.com/ivmai/bdwgc/tree/master/doc github by default displays "readme.autoconf" which is fairly confusing...

Ah, it's surprise to me.

> maybe I could add a README.md there that says "see overview.md for an explanation of the documentation files here"?

At this moment I am not OK with adding another readme.md for this purpose. I hope newcomers start with the root readme. I don't have a good idea right now how to fix it. Thank you for pointing me to this issue anyway.

PR #519, opened 2022-12-15
Add support for building with wasm32-wasi instead of emscripten

anuraaga:
I am using bdwgc in a webassembly program that is built directly with LLVM without emscripten - this uses `wasm32-wasi` and wasi-sdk as opposed to `wasm32-unknown`. For the most part, behavior is similar such as lack of threads and the Wasm page size. Some differences are that header files for non-supported features aren't available at all (I had to add a guard to `setjmp.h` which seems possible with an existing macro) and there are symbols for setting the location of the data segment (I think this may actually be the same for emscripten but not sure, it seems to be currently set to just `ALIGNMENT` and I didn't try changing that in this PR). 

For reference, I use this patch to build bdwgc for inclusion in a wasm project here

https://github.com/corazawaf/coraza-proxy-wasm/blob/main/buildtools/bdwgc/Dockerfile

One thing noticable in the Dockerfile was I was able to work around the lack of signals with `-D_WASI_EMULATED_SIGNAL`, but a more precise approach would be guarding the `signal.h` includes. This would mean bringing back a previously deleted `NO_SIGNALS` macro though

### DIFF
--- a/include/private/gc_priv.h
+++ b/include/private/gc_priv.h
@@ -857,7 +857,9 @@ EXTERN_C_END
 # endif
 #endif /* DARWIN */
 
-#include <setjmp.h>
+#ifndef GC_NO_SIGSETJMP
+#  include <setjmp.h>
+#endif
 
 #include <stdio.h>
 
--- a/include/private/gc_priv.h
+++ b/include/private/gc_priv.h
@@ -857,7 +857,7 @@ EXTERN_C_END
 # endif
 #endif /* DARWIN */
 
-#ifndef WASI
+#ifndef WASI /* wasi does not provide this header */
 #  include <setjmp.h>
 #endif
 
--- a/include/private/gc_priv.h
+++ b/include/private/gc_priv.h
@@ -857,7 +857,7 @@ EXTERN_C_END
 # endif
 #endif /* DARWIN */
 
-#ifndef GC_NO_SIGSETJMP
+#ifndef WASI
 #  include <setjmp.h>
 #endif
 
--- a/include/private/gcconfig.h
+++ b/include/private/gcconfig.h
@@ -1319,9 +1319,9 @@ EXTERN_C_BEGIN
 #   endif
 #   ifdef WASI
 #     define OS_TYPE "WASI"
-      extern unsigned char __global_base;
+      extern char __global_base;
 #     define DATASTART ((ptr_t)&__global_base)
-      extern unsigned char __heap_base;
+      extern char __heap_base;
 #     define DATAEND ((ptr_t)&__heap_base)
       /* The real page size in WebAssembly is 64 KB.    */
 #     define GETPAGESIZE() 65536
--- a/include/private/gcconfig.h
+++ b/include/private/gcconfig.h
@@ -484,6 +484,15 @@ EXTERN_C_BEGIN
 #   endif
 #   define I386
 #   define mach_type_known
+#   define WASM
+# endif
+# if defined(__wasi__) || defined(WASI)
+#   ifndef WASI
+#     define WASI
+#   endif
+#   define I386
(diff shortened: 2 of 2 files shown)


include/private/gc_priv.h
```diff
@@ -857,7 +857,9 @@ EXTERN_C_END
 # endif
 #endif /* DARWIN */
 
-#include <setjmp.h>
+#ifndef GC_NO_SIGSETJMP
```
ivmai, line 860: No, this is not correct - setjmp.h is included to have setjmp() which is used if no sigsetjmp(). It seems we need ifndef WASI

include/private/gcconfig.h
```diff
@@ -484,6 +484,15 @@ EXTERN_C_BEGIN
 #   endif
 #   define I386
 #   define mach_type_known
+#   define WASM
```
ivmai, line 487: Defining here something else than <ARCH> or <OS> is not recommended. Even if it leads to some code duplication.

include/private/gcconfig.h
```diff
@@ -484,6 +484,15 @@ EXTERN_C_BEGIN
...
+# endif
+# if defined(__wasi__) || defined(WASI)
+#   ifndef WASI
```
ivmai, line 490: This seems to be  copy paste of Emscripten code. Could we change it to?
```
# if defined(__wasi__)
# define WASI
...
```

include/private/gcconfig.h
```diff
@@ -1310,6 +1316,23 @@ EXTERN_C_BEGIN
 #     if defined(GC_THREADS) && !defined(CPPCHECK)
 #       error No threads support yet
 #     endif
+#     ifdef EMSCRIPTEN
+#       define OS_TYPE "EMSCRIPTEN"
+#       define DATASTART (ptr_t)ALIGNMENT
+#       define DATAEND (ptr_t)ALIGNMENT
+#     endif
+#     ifdef WASI
+        extern unsigned char __global_base;
```
ivmai, line 1325: I think "unsigned" is redundant here. Also, please keep them more close to DATASTART/END

include/private/gc_priv.h
anuraaga, line 860: It looks like my hypothesis that using this macro here would be safe was invalid

https://ci.appveyor.com/project/ivmai/bdwgc/builds/45688100/job/72kkvn3k4tfdwxro

Not too sure what would be a good way to handle this missing header file though, appreciate any pointer

include/private/gcconfig.h
```diff
@@ -1310,6 +1316,23 @@ EXTERN_C_BEGIN
...
+        extern unsigned char __heap_base;
+#       define OS_TYPE "WASI"
+#       define GETPAGESIZE() 65536
```
ivmai, line 1328: I think we need to add comment that for GETPAGESIZE same as in NACL case.

ivmai (changes requested):

ivmai:
> I think this may actually be the same for emscripten but not sure, it seems to be currently set to just ALIGNMENT and I didn't try changing that in this PR

I can't comment on this, probably @juj could.

anuraaga:
Thanks for the quick review! Addressed the comments I think

> I think this may actually be the same for emscripten but not sure, it seems to be currently set to just ALIGNMENT and I didn't try changing that in this PR
> I can't comment on this, probably @juj could.

Now that the config block isn't shared at all, it's probably not a big deal either way

ivmai:
Thank you. Committed (with some adjustment of comments) as fbc4593

anuraaga:
Thanks @ivmai! Can confirm it works well, no more patching https://github.com/corazawaf/coraza-proxy-wasm/pull/119/files

PR #520, opened 2022-12-15
Update README.cmake

rdp:
Add description of how to install via cmake.


ivmai:
Thank you. Committed with come changes as a6b6720

ivmai:
BTW. Just if you have time and will, it would be better to convert this readme (as well as other non-platform readme files) to .md

rdp:
Thanks.  I'm not exactly sure how to convert them, or are you thinking just the filenames?

ivmai:
Yes, the minimum is to rename (remove "README.", add .md) and update links. The rest is a creative work to make it better readable if viewed on github - to insert formatting symbols where appropriate (like `code`) and convert to links if other doc files are mentioned in these files

PR #521, opened 2022-12-15
Update README.macros

rdp:
Try to define difference between include time macros and build-time settings better.


ivmai:
Thank you committed as c6e1ae1 (with some more changes in the file)

PR #527, opened 2023-02-20
Fix typo in scale.md

deadalnix:
As per title.


ivmai:
Thank you for catching this! Committed as 3d7cc9be4

PR #504, opened 2022-11-21
Simplify obtaining the stack base for Emscripten

juj:
Simplify obtaining the stack base for Emscripten by using the direct function `emscripten_stack_get_base()`, instead of the ASYNCIFY-only function `emscripten_scan_stack()`.

PR #530, opened 2023-03-02
Emscripten: Remove use of `emscripten_scan_stack()`.

waywardmonkeys:
The code was only enabled by a define and not needed as the alternative is to just use `emscripten_stack_get_base()` which is more straight forward and avoids an extra file-scope static.

There was nothing controlling the define enabling this.

### DIFF
--- a/os_dep.c
+++ b/os_dep.c
@@ -1199,31 +1199,10 @@ GC_INNER size_t GC_page_size = 0;
   }
 # define GET_MAIN_STACKBASE_SPECIAL
 #elif defined(EMSCRIPTEN)
-
-# if defined(USE_EMSCRIPTEN_SCAN_STACK) && defined(EMSCRIPTEN_ASYNCIFY)
-    /* According to the documentation, emscripten_scan_stack() is only  */
-    /* guaranteed to be available when building with ASYNCIFY.          */
-#   include <emscripten.h>
-
-    static void *emscripten_stack_base;
-
-    static void scan_stack_cb(void *begin, void *end)
-    {
-      (void)begin;
-      emscripten_stack_base = end;
-    }
-# else
 #   include <emscripten/stack.h>
-# endif
-
   ptr_t GC_get_main_stack_base(void)
   {
-#   if defined(USE_EMSCRIPTEN_SCAN_STACK) && defined(EMSCRIPTEN_ASYNCIFY)
-      emscripten_scan_stack(scan_stack_cb);
-      return (ptr_t)emscripten_stack_base;
-#   else
       return (ptr_t)emscripten_stack_get_base();
-#   endif
   }
 # define GET_MAIN_STACKBASE_SPECIAL
 #elif !defined(AMIGA) && !defined(HAIKU) && !defined(OS2) \


waywardmonkeys:
FYI @juj 

Also, I'm unable to successfully run the emscripten port successfully yet.

ivmai:
The code was developed by @pclouds (we'd better ask the author why emscripten_scan_stack was preferred over emscripten_stack_get_base).

> There was nothing controlling the define enabling this.

CFLAGS is controlling it, the one could compile passing `-D USE_EMSCRIPTEN_SCAN_STACK`.

> Also, I'm unable to successfully run the emscripten port successfully yet.

At this moment I don't think it is the right time to drop that code. Let's first make the port works reliably on Emscripten.
I guess, probably you are right that code under USE_EMSCRIPTEN_SCAN_STACK does not give any preference, may be, but not now. Thank you for pointing it out, anyway!

waywardmonkeys:
> The code was developed by @pclouds (we'd better ask the author why emscripten_scan_stack was preferred over emscripten_stack_get_base).

The code was originally written with `emscripten_scan_stack()`, but then switched to `emscripten_stack_get_base()` as `emscripten_scan_stack()` was documented to only work when `ASYNCIFY` is enabled. (It ends up that that is actually wrong in the Emscripten documentation, so I am working through a series of PRs to fix the docs on the Emscripten side as well.)

All the function does on the emscripten side is this:

```
void emscripten_scan_stack(em_scan_func func) {
  uintptr_t base = emscripten_stack_get_base();
  uintptr_t end = emscripten_stack_get_current();
  func((void*)end, (void*)base);
}
```

So the earlier change to just use `emscripten_stack_get_base()` directly was the correct thing to do.

(The earlier changes were done as part of #504 / e4247e8e31621563d53af29910d51ffe1427d3f2.)

> > There was nothing controlling the define enabling this.
> 
> CFLAGS is controlling it, the one could compile passing `-D USE_EMSCRIPTEN_SCAN_STACK`.

I just mean that the existing build configurations don't need to toggle this as there's no need for the code.

> > Also, I'm unable to successfully run the emscripten port successfully yet.
> 
> At this moment I don't think it is the right time to drop that code. Let's first make the port works reliably on Emscripten. I guess, probably you are right that code under USE_EMSCRIPTEN_SCAN_STACK does not give any preference, may be, but not now. Thank you for pointing it out, anyway!

I'm working on further patches ... but wanted standalone changes that are easy to review and understand in isolation.

I'd like to  keep this open so that I can layer subsequent changes without too much gymnastics.

(I'm also porting another GC to WebAssembly, which was why I started looking at this to see how they compared.)

ivmai:
> CFLAGS is controlling it, the one could compile passing -D USE_EMSCRIPTEN_SCAN_STACK.

Aha, sorry, I'm wrong here.

ivmai:
> I'm working on further patches ... but wanted standalone changes that are easy to review and understand in isolation.

So, you think it would be better to apply this patch now?

waywardmonkeys:
> So, you think it would be better to apply this patch now?

I do!

Maybe @juj will chime in … @pclouds hasn’t been active in over a year.

ivmai:
Okay, I will apply it today or tomorrow

pclouds:
I'm still here :) even though I don't have much to contribute. I more or less gave up on my attempt to make gauche scheme work on wasm (the reason I looked at bdwgc). If you think this is a good way to do it, go ahead

ivmai:
Thank you, @pclouds

juj:
This looks like a good simplification, indeed just using `emscripten_stack_get_base()` will work with both with ASYNCIFY and without.

PR #532, opened 2023-03-03
Emscripten, gctest: Remove EXPORTED_FUNCTIONS.

waywardmonkeys:
When `mmap` was used, there was apparently a need to export `memalign` and `memset` from the WebAssembly code to JS for the mmap emulation code to use.

The `mmap` code path is no longer used with the Emscripten support, so this export is no longer needed.

The explicit export of `_main` is not required. According to the Emscripten code:

    By default if this setting is not specified on the command
    line the `_main` function will be implicitly exported.

* tests/tests.am [EMSCRIPTEN]: Remove `gctest_html_LDFLAGS`.

### DIFF
--- a/tests/tests.am
+++ b/tests/tests.am
@@ -30,9 +30,6 @@ if EMSCRIPTEN
 check_PROGRAMS += gctest.html
 gctest_html_SOURCES = $(gctest_SOURCES)
 gctest_html_LDADD = $(gctest_LDADD)
-# Bug in the linker not being able to determine that _memalign and _memset
-# are needed? it's part of mmap.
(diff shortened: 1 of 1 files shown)


waywardmonkeys:
FYI @juj, @pclouds

ivmai:
BTW. If possible please also check the build for Emscripten with CMake script

waywardmonkeys:
Sure. This particular one isn’t going to impact that as this in the automaker config… but other PRs could need it.

At some point, I plan to ask about adding a GutHub Action or three that test some of these configurations.

ivmai:
> At some point, I plan to ask about adding a GutHub Action or three that test some of these configurations.

Okay.

PR #534, opened 2023-03-05
WebAssembly: Get page size from "OS".

waywardmonkeys:
Instead of defining `GETPAGESIZE()` for WASI, we can do what we do in Emscripten and other targets and just fall back to the default handling of this which invokes `getpagesize()`.

* include/private/gcconfig.h [WEBASSEMBLY + EMSCRIPTEN]: Remove dangling comment.
* include/private/gcconfig.h [WEBASSEMBLY WASI] (GETPAGESIZE): Remove.

### DIFF
--- a/include/private/gcconfig.h
+++ b/include/private/gcconfig.h
@@ -2384,7 +2384,6 @@ EXTERN_C_BEGIN
       /* native support of memory mapping.  Use sbrk() instead.     */
 #     undef USE_MMAP
 #     undef USE_MUNMAP
-      /* The real page size in WebAssembly is 64 KB.    */
 #     if defined(GC_THREADS) && !defined(CPPCHECK)
 #       error No threads support yet
 #     endif
@@ -2403,8 +2402,6 @@ EXTERN_C_BEGIN
 #     endif
 #     undef USE_MMAP /* similar to Emscripten */
 #     undef USE_MUNMAP
-      /* The real page size in WebAssembly is 64 KB.    */
-#     define GETPAGESIZE() 65536
 #     if defined(GC_THREADS) && !defined(CPPCHECK)
 #       error No threads support yet
 #     endif


waywardmonkeys:
Both Emscripten and WASI implement the required libc API, so this hardcoding here shouldn't be necessary. I confirmed that the same thing gets set before and after this change.

cc: @juj @anuraaga

anuraaga (approved):

juj:
If I interpret this correctly, removing these lines would make bdwgc call the `Linux function getpagesize()`(https://man7.org/linux/man-pages/man2/getpagesize.2.html) at runtime?

Looking at that page, there are mentions

```
In SUSv2 the getpagesize() call is labeled LEGACY, and in POSIX.1-2001 it has been dropped; HP-UX does not have this call.
```
and
```
Portable applications should employ sysconf(_SC_PAGESIZE) instead of getpagesize()
```

However I would recommend avoiding this kind of direction in general, where Emscripten would be seen as an OS and POSIXisms would be used to access the OS layer.

Hardcoding the WebAssembly page size directly into a header has a benefit of providing a compile time literal, which will automatically fold away to all calling code at compile time. Calling a runtime function like `sysconf(_SC_PAGESIZE)` will require a LTO optimizing inliner for the same result to potentially happen. If people are not using `-flto` in their Emscripten builds, they will get increased code size in their bdwgc Wasm compiles.

The WebAssembly standard has the 64KB page size rooted deep in its specification and [user-facing API](https://developer.mozilla.org/en-US/docs/WebAssembly/JavaScript_interface/Memory/grow), so it is not possible for it to transparently change as an implementation detail in any future.

Therefore I would recommend that continuing to hardcode this value will provide the best possible code size out of the box. I don't think hardcoding per se is bad.

Looking at https://github.com/emscripten-core/emscripten/blob/667505546e57040aaf00a56c07c0b924158307e9/system/lib/libc/musl/include/limits.h#L97 , it may be possible that calling

```
#include <limits.h>
...
int pageSize = PAGE_SIZE;
```
might give the WebAssembly page size as a compile-time literal, if it is uncomfortable for bdwgc repository to contain a direct number value for the page size. If that does not work, then a second alternative is to call

```
#include <emscripten/heap.h>
...
int pageSize = WASM_PAGE_SIZE;
```
from https://github.com/emscripten-core/emscripten/blob/667505546e57040aaf00a56c07c0b924158307e9/system/include/emscripten/heap.h#L16

anuraaga (changes requested):
Just to invalidate review as above

ivmai:
> include/private/gcconfig.h [WEBASSEMBLY + EMSCRIPTEN]: Remove dangling comment.

No, it is not dangling, I intentionally retained it to indicate that the page size is much bigger than HBLKSIZE.
OK, to delete it, of course.

ivmai:
> Hardcoding the WebAssembly page size directly into a header has a benefit of providing a compile time literal, which will automatically fold away to all calling code at compile time.

I think this is just nominal benefit, not visible in practice. Correct me if I'm wrong.

> The WebAssembly standard has the 64KB page size rooted deep in its specification and [user-facing API](https://developer.mozilla.org/en-US/docs/WebAssembly/JavaScript_interface/Memory/grow), so it is not possible for it to transparently change as an implementation detail in any future.

Okay, let's retain GETPAGESIZE() as is.

@waywardmonkeys, please let me drop this patch, I don't see a reason enough for a change (but, I should point out that your suggestion has some ground). Thank you anyway!

ivmai:
Okay, let's apply this patch. As some refactoring.

juj:
> The WASI build hardcodes GETPAGESIZE(), while Emscripten is falling back on getpagesize(). Why should one be different from the other?

I am sorry if there was some confusion. I never argued there would be a need for these two to be different, the reason they currently are seems to be historical (different developers contributed to Emscripten support and WASI support in bdwgc).

Hardcoding 64KB for both would be fine.

> The argument about a function call isn't entirely accurate since this is something that happens when GC_setpagesize() is called. (Usually during GC_init().)

If I understand the implied meaning of this sentence correctly - the recommendation I posted was not about hotness or coldness of `getpagesize()/sysconf()`. For purposes of code disk size, it does not matter how many or few times the function is called, but the code size grows already at the presence of the first call, and an LTO optimizer is then needed to fold the function away.

For example, the POSIX function `sysconf()` is emulated in Emscripten, and it is quite a large switch-case:

https://github.com/emscripten-core/emscripten/blob/0ec03510481b70c8980c947ae398d0ea83ae5ad0/system/lib/libc/musl/src/conf/sysconf.c#L30-L232

It would be nice not to include that kind of a construct in the build for "nothing". Emscripten is a special platform compared to all the rest in that it has to download all the compiled bytes live for the page to start up, so every byte can count, and this type of "bloat" is accumulated from small streams.

That is why I recommended just hardcoding the value, since in this case it will be free to do so, the spec provides this guarantee, and doing so will provide the best code size in debug and release builds.

Best practices for targeting the web and Emscripten are good to avoid POSIXisms that do not map natively to the web. We did work hard for almost a decade to add POSIX build cross-compatibility in Emscripten, but then learned that in general terms such native APIs are not a direct fit to the web, and smaller code is gained by providing "native" web abstractions instead that embrace the kind of "native webby" primitives.

Anyhow, in this case it is not a big deal either way, just a micro-optimization.

PR #503, opened 2022-11-21
Don't attempt multithreaded Emscripten builds

juj:
If I understand correctly, Boehm requires being able to stop all managed threads before being able to collect. (via e.g. pthread_stop_world.c)? (are there any alternative solutions?)

WebAssembly/Web Workers do not currently/yet support synchronous stopping of all threads, so disallow building in multithreaded mode with Emscripten, and hint the user to build singlethreaded Bdwgc instead.

For future reference, attempting to build to WebAssembly with multithreading enabled currently fails with

```
[28/28] Linking C executable gctest.js
FAILED: gctest.js
cmd.exe /C "cd . && C:\emsdk\emscripten\main\emcc.bat -g  CMakeFiles/gctest.dir/tests/gctest.c.o -o gctest.js  libgc.a && cd ."
error: undefined symbol: pthread_kill (referenced by top-level compiled C/C++ code)
warning: Link with `-sLLD_REPORT_UNDEFINED` to get more information on undefined symbols
warning: To disable errors for undefined symbols use `-sERROR_ON_UNDEFINED_SYMBOLS=0`
warning: _pthread_kill may need to be added to EXPORTED_FUNCTIONS if it arrives from a system library
error: undefined symbol: sem_getvalue (referenced by top-level compiled C/C++ code)
warning: _sem_getvalue may need to be added to EXPORTED_FUNCTIONS if it arrives from a system library
error: undefined symbol: sigsuspend (referenced by top-level compiled C/C++ code)
warning: _sigsuspend may need to be added to EXPORTED_FUNCTIONS if it arrives from a system library
Error: Aborting compilation due to previous errors
emcc: error: 'C:/emsdk/node/14.18.2_64bit/bin/node.exe C:\emsdk\emscripten\main\src\compiler.js C:\Users\clb\AppData\Local\Temp\tmpkxx2t1u5.json' failed (returned 1)
ninja: build stopped: subcommand failed.
```

PR #535, opened 2023-03-05
WebAssembly+cmake: Default to enable_threads=OFF.

waywardmonkeys:
Since the GC doesn't support threading when under Emscripten or WASI, we should default to building without threads.

This was tested with cmake in 3 ways:

* `cmake -GNinja ..` on Linux. Threads remained enabled.
* `emcmake cmake -GNinja ..` with the Emscripten SDK. Threads defaulted to off.
* `cmake -DCMAKE_TOOLCHAIN_FILE=/usr/share/cmake/wasi-sdk.cmake -GNinja ..` with the wasi-sdk under Docker. Threads defaulted to off. (This build had other issues that need to be sorted out.)

If this is approved, I will do one for `enable_mmap` and `enable_munmap` ... or I could add it to this PR as well.
cc: @juj @anuraaga

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -59,12 +59,17 @@ if (NOT (${CMAKE_VERSION} VERSION_LESS "3.18.0"))
   include(CheckLinkerFlag)
 endif()
 
+set(default_enable_threads ON)
+if(EMSCRIPTEN OR WASI)
+  set(default_enable_threads OFF)
+endif()
+
 # Customize the build by passing "-D<option_name>=ON|OFF" in the command line.
 option(BUILD_SHARED_LIBS "Build shared libraries" ON)
 option(build_cord "Build cord library" ON)
 option(build_tests "Build tests" OFF)
 option(enable_docs "Build and install documentation" ON)
-option(enable_threads "Support threads" ON)
+option(enable_threads "Support threads" ${default_enable_threads})
 option(enable_parallel_mark "Parallelize marking and free list construction" ON)
 option(enable_thread_local_alloc "Turn on thread-local allocation optimization" ON)
 option(enable_threads_discovery "Enable threads discovery in GC" ON)


anuraaga (approved):

juj:
This looks great to me 👍

ivmai:
No, I don't like it. By default, the build is multithreaded for any platform. If the platform or build variant is not supported then just fail.
Cmake and configure - the build is multithreaded by default.
Makefile.direct - the build is single-threaded by default.
I'd like not to touch this logic. Is it Okay?

anuraaga:
I'm not sure about emscripten but at least with wasi-sdk, the build fails obviously due to lack of pthread, which is pretty obvious just then disable it. For me it wasn't a bad experience so I'd be ok with maintaining that stance.

waywardmonkeys:
I found it annoying that I had to look up which flags to use to get a working build ... but that could also be fixed by improving the documentation instead.

But whatever is decided, I'm fine.

ivmai:
>  configure - the build is multithreaded by default.

This is not correct, the proper is: configure - the build is multithreaded by default if threads are supported.

It looks like I got your point - match the behavior of configure. But can we do w/o hardcoding Emscripted/Wasi in cmake script?

ivmai:
I mean by using   find_package(Threads)

ivmai:
The logic of configure is: if there is not --enable/disable-threads option given, then turn on threads if threads are supported. We cannot exactly reproduce this behavior in cmake as I understand but let's try to do something close to.

ivmai:
Can we do like this?
Define the option as `option(enable_threads "Support threads" AUTO)` and then if AUTO then decide on/off automatically, otherwise the option is assumed to be ON or OFF. @waywardmonkeys, could you please try it?

juj:
> Define the option as option(enable_threads "Support threads" AUTO) and then if AUTO then decide on/off automatically

On the web it is actually not possible to decide at runtime if multithreading is supported or not, but the decision to target multithreading must be chosen at compilation time. This is because it affects the web hosting that one needs to set up, and the minimum WebAssembly feature version that will be targeted.

Also Bdwgc currently is unable to work on WebAssembly with multithreading enabled due to Wasm standard limitations, which is why it would be appealing to default to having threads disabled.

ivmai:
> This was tested with cmake in 3 ways:

BTW. It would be nice to add at least one config to .travis.yml for Emscripten or wasi testing.

CMakeLists.txt
```diff
@@ -59,12 +59,17 @@ if (NOT (${CMAKE_VERSION} VERSION_LESS "3.18.0"))
   include(CheckLinkerFlag)
 endif()
 
+set(default_enable_threads ON)
+if(EMSCRIPTEN OR WASI)
```
ivmai, line 63: How about change this to?
+find_package(Threads QUIET)
+if (NOT (CMAKE_USE_PTHREADS_INIT OR CMAKE_USE_WIN32_THREADS_INIT) OR EMSCRIPTEN OR WASI)

ivmai: I have modified it. Please check commit [f21eefa](https://github.com/ivmai/bdwgc/commit/f21eefafc047240a4d971de5147bf64dd46725fe)

ivmai:
Please test commit f21eefa on Emscripten or WASI

ivmai:
Related issue #503

PR #551, opened 2023-05-07
Add hurd x86_64 support

sthibaul:


ivmai:
Committed as 83c84ab96. Thank you

PR #566, opened 2023-08-26
Fix a bug where objects are reclaimed too early with multiple threads on Windows

spaghettisalat:
The issue arises because some registers that can contain pointers given the right optimization settings were not searched during the marking phase. I encountered this on Windows x86_64 using gcc 13.2.0. For `-march=native` on reasonably recent hardware, gcc autovectorizes code like

```
struct test_struct {
  void *a;
  void *b;
};

struct test_struct *alloc_test_struct(void *a, void *b) {
  struct test_struct *obj;
  obj = GC_MALLOC(sizeof(struct test_struct));
  obj->a = a;
  obj->b = b;
  return obj;
}
```

such that the pointers `a` and `b` are stored in the `xmm6` register. Since this register wasn't searched, the pointers where invisible to the gc and could be reclaimed too early.

A simple test program showing the bug can be found at https://github.com/spaghettisalat/bdwgc-test-program. If necessary, I can add this as a test to the bdwgc repository.

This bug might also affect other architectures, but since I don't have access to any Windows machines running those this merge request only touches the x86_64 dependent parts.

### DIFF
--- a/win32_threads.c
+++ b/win32_threads.c
@@ -379,6 +379,8 @@ GC_INLINE LONG GC_get_max_thread_index(void)
                         ? CONTEXT_INTEGER | CONTEXT_CONTROL \
                           | CONTEXT_EXCEPTION_REQUEST | CONTEXT_SEGMENTS \
                         : CONTEXT_INTEGER | CONTEXT_CONTROL)
+#elif defined(X86_64)
+# define GET_THREAD_CONTEXT_FLAGS (CONTEXT_INTEGER | CONTEXT_CONTROL | CONTEXT_FLOATING_POINT)
 #else
 # define GET_THREAD_CONTEXT_FLAGS (CONTEXT_INTEGER | CONTEXT_CONTROL)
 #endif /* !WOW64_THREAD_CONTEXT_WORKAROUND */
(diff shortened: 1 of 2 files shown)


win32_threads.c
```diff
@@ -377,10 +377,10 @@ GC_INLINE LONG GC_get_max_thread_index(void)
   static GC_bool isWow64; /* Is running 32-bit code on Win64?   */
 # define GET_THREAD_CONTEXT_FLAGS (isWow64 \
                         ? CONTEXT_INTEGER | CONTEXT_CONTROL \
-                          | CONTEXT_EXCEPTION_REQUEST | CONTEXT_SEGMENTS \
-                        : CONTEXT_INTEGER | CONTEXT_CONTROL)
+                          | CONTEXT_EXCEPTION_REQUEST | CONTEXT_SEGMENTS | CONTEXT_FLOATING_POINT\
```
ivmai, line 380: Please revert these 2 lines (the patch is not about 32-bit mode)

win32_threads.c
```diff
@@ -685,6 +685,14 @@ static ptr_t copy_ptr_regs(word *regs, const CONTEXT *pcontext) {
 #   elif defined(X86_64)
       PUSH4(Rax,Rcx,Rdx,Rbx); PUSH2(Rbp, Rsi); PUSH1(Rdi);
       PUSH4(R8, R9, R10, R11); PUSH4(R12, R13, R14, R15);
+      PUSH4(Xmm0.Low, Xmm0.High, Xmm1.Low, Xmm1.High);
```
ivmai, line 688: I have a question: do we need to save High parts too?

ivmai:
It looks like even ancient x64 chips (supported by Windows) had XMM registers, so we don't need to add any runtime CPU checks.

win32_threads.c
ivmai: https://newbedev.com/128-bit-values-from-xmm-registers-to-general-purpose says SSE4.1 has pextrq to extract High part of xmm but does gcc/clang practically use it?

win32_threads.c
ivmai: It should look like this:
```
#ifdef WOW64_THREAD_CONTEXT_WORKAROUND
... // no change
#elif defined(I386)
# define GET_THREAD_CONTEXT_FLAGS (CONTEXT_INTEGER | CONTEXT_CONTROL)
#else
# define GET_THREAD_CONTEXT_FLAGS (CONTEXT_INTEGER | CONTEXT_CONTROL |  CONTEXT_FLOATING_POINT)
#endif /* X86_64 */
```

ivmai:
> This bug might also affect other architectures, but since I don't have access to any Windows machines running those this merge request only touches the x86_64 dependent parts.

Okay, let's fix it for x64 for now

win32_threads.c
spaghettisalat: > Please revert these 2 lines (the patch is not about 32-bit mode)

Done

win32_threads.c
spaghettisalat: Yes, gcc also uses the high parts. In the example above, the two pointers `a` and `b` are stored in the low and high parts of the register respectively.

ivmai:
Thank you. I've merged your code (commit c245cf6) with minor modifications (e.g. I've added a macro XMM_CANT_STORE_PTRS to prevent storing of xmm register if the client could guarantee that no pointers are xmm registers).
Please check.

PR #574, opened 2023-09-15
Use page size for WASI

anuraaga:
I tried updating bdwgc in my WASI build after some time and ran into an issue where sbrk isn't called with page size increments, which is required by wasi

https://github.com/WebAssembly/wasi-libc/blame/ec4566beae84e54952637f0bf61bee4b4cacc087/libc-bottom-half/sources/sbrk.c#L14

This wasn't a problem before so I imagine some code got changed or refactored to affect this, so I added a guard for WASI where it's controlled whether real or fake page size is used.


ivmai:
Committed as [f8bfbfe](https://github.com/ivmai/bdwgc/commit/f8bfbfe965cea9c17798132e35c41d727436487b).
Thank you

PR #576, opened 2023-09-23
[pthread_support] moving `first_thread_used` to the global Scope

catap:
In this change, we relocate the `first_thread_used` variable to the global scope. This adjustment is made to avoid potential issues that may arise when defining a variable as `static` on the stack in applications with parallel access to such variables.

When a variable is declared as `static` on the stack, compilers typically use a hidden flag to indicate whether the static variable has been initialized. This flag is checked each time the function is entered. While this incurs a minor performance overhead, a more significant concern is the lack of a guarantee that this flag is thread-safe or atomic.

When static variables on the stack are accessed by multiple threads, race conditions can occur, leading to incorrect or even multiple initializations of the variable. Such situations may result in overwriting pointers inside `GC_threads` or `first_thread`, potentially triggering memory cleanup by the GC cycle and causing unforeseen issues to arise.

Notable to say that win32 threads hasn't got that issue.

### DIFF
--- a/pthread_support.c
+++ b/pthread_support.c
@@ -618,6 +618,7 @@ GC_INNER GC_thread GC_threads[THREAD_TABLE_SZ] = {0};
 /* other fields (crtn) to be pushed.                                    */
 static struct GC_StackContext_Rep first_crtn;
 static struct GC_Thread_Rep first_thread;
+static GC_bool first_thread_used = FALSE;
 
 /* A place to retain a pointer to an allocated object while a thread    */
 /* registration is ongoing.  Protected by the GC lock.                  */


ivmai:
Sorry, I don't understand your explanation as well as the code change.
What is "static on the stack"?
Could you please point me to some documentation about it?

Also, have you observed race in first_thread actually?

catap:
With this patch I can't reproduce https://gitlab.com/embeddable-common-lisp/ecl/-/issues/718 on my laptop. It isn't a fix but it dramatically decrease probability. So, I assume that "some" symptoms of race condition in `first_thread`.

As `static on the stack` I mean a local variable on the function stack which is defined with `static` keyword. For example:
```
void foo() {
  static int a = 1;
  ...
}
```

ivmai:
It would be nice if you compare the asm code before vs after the change. In my opinion it should not change because the initialization value is constant, but may be I am mistaken

ivmai:
> It would be nice if you compare the asm code before vs after the change. In my opinion it should not change because the initialization value is constant, but may be I am mistaken

Maybe it is compiler-dependant, so please check

catap:
@ivmai tricky part that ASM code will be the same as you expected as you said ;)

I've googled a bit and discovered the thread on stackoverflow about this behaviour: https://stackoverflow.com/questions/246564/what-is-the-lifetime-of-a-static-variable-in-a-c-function

ivmai:
> With this patch I can't reproduce https://gitlab.com/embeddable-common-lisp/ecl/-/issues/718 on my laptop. It isn't a fix but it dramatically decrease probability. So, I assume that "some" symptoms of race condition in `first_thread`.
> 
> As `static on the stack` I mean a local variable on the function stack which is defined with `static` keyword. For example:
> 
> ```
> void foo() {
>   static int a = 1;
>   ...
> }
> ```

Probably you could add some dummy static variable and have the same behavior (dramatical decrease of the likelyhood of the error).

ivmai:
Let's reject this PR?

PR #586, opened 2023-11-21
*_allocator<void>: Make members public

roberth:
A private value_type is a problem for boost small_vector<X>, which instantiates rebind<void> at some point.

    include/boost/container/allocator_traits.hpp:145:37: error: 'value_type' is a private member of 'traceable_allocator<void>'
       typedef typename allocator_type::value_type value_type;

[...]

    gc_allocator.h:319:23: note: implicitly declared private here
      typedef void        value_type;

I don't see a reason why any of the members should be private, and arguably it violates the Liskov substitution principle. It seems that it was left private by accidental omission.

### DIFF
--- a/include/gc/gc_allocator.h
+++ b/include/gc/gc_allocator.h
@@ -170,6 +170,7 @@ public:
 
 template<>
 class gc_allocator<void> {
+public:
   typedef GC_ALCTR_SIZE_T    size_type;
   typedef GC_ALCTR_PTRDIFF_T difference_type;
   typedef void*       pointer;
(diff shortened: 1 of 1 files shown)


ivmai:
Agree. But I think same should be done for gc_allocator<void> and gc_allocator_ignore_off_page<void>

roberth:
Good point; done.

Perhaps superfluously, I think the CI failures are unrelated (no C++ involved) and flaky (went from 4 to 2 failures for no apparent reason)

ivmai:
Merged

Issue #596, opened 2023-12-13
Regression: support of macOS before 10.7


catap:
Since https://github.com/ivmai/bdwgc/commit/18b511147d818d37824826b5d8cffa3a73cf91ba it is mandatory to use `getsectiondata` on macOS out of the box.

Unfortunately this system call was introduced at 10.7: see https://github.com/phracker/MacOSX-SDKs/blob/master/MacOSX10.6.sdk/usr/include/mach-o/getsect.h vs https://github.com/phracker/MacOSX-SDKs/blob/master/MacOSX10.7.sdk/usr/include/mach-o/getsect.h

The right fix is track support of `getsectiondata` inside configure and enforce `USE_GETSECTBYNAME` when `getsectiondata` isn't available.

ivmai:
I see. I'll review the patch.

ivmai:
> The right fix is track support of getsectiondata inside configure and enforce USE_GETSECTBYNAME when getsectiondata isn't available.

It might look as the right fix, I understand, but there is a set of build scripts in bdwgc and I'm a bit resistant against moving checks to the build scripts. Looking into the bdwgc source, we may see the use of [AvailabilityMacros.h](https://opensource.apple.com/source/CarbonHeaders/CarbonHeaders-8A428/AvailabilityMacros.h.auto.html) at some places.
How about the following fix, could you check it, please?
Add the following 3 lines at https://github.com/ivmai/bdwgc/blob/98b5d3895fff0c3abdcb8d8e46d085198e6e15f8/dyn_load.c#L1316
```
#if MAC_OS_X_VERSION_MAX_ALLOWED <= 1200 /*MAC_OS_X_VERSION_12_0*/
# define USE_GETSECTBYNAME
#endif
```

ivmai:
@coolbikerdad Could you please check the above 3-lines fix too?

catap:
@ivmai sure, it also works.

coolbikerdad:
> @coolbikerdad Could you please check the above 3-lines fix too?

On my macOS 14 system this is OK.

catap:
> > @coolbikerdad Could you please check the above 3-lines fix too?
> 
> On my macOS 14 system this is OK.

this is regression only for *very* old macOS. For everything before Lion: https://en.wikipedia.org/wiki/OS_X_Lion which was released in 2011.

ivmai:
> \#if MAC_OS_X_VERSION_MAX_ALLOWED <= 1200
[...]

PR #597, opened 2023-12-13
Enforce `USE_GETSECTBYNAME` on macOS before 10.7

catap:
Fixes https://github.com/ivmai/bdwgc/issues/596


catap:
@ivmai changed fix as you requested

ivmai:
Ah, I did it in parallel (commit 0a6e4c335). Slightly different, but should not be an issue.
Thank you for reporting and collaborating on the issue!

PR #603, opened 2023-12-28
Add constexpr support to allocators for C++20

jeaye:
This adds the `constexpr` attribute on each allocator member function, including ctors and dtors, to enable easier use of these allocators in compile-time execution. A practical example of this would be a modern C++ string class which supports constexpr construction, usage, and destruction. Without this change, the C++ compiler will not allow such a class.

An trivial example to show the usage:

```cpp
// Compile with: c++ -std=c++20 -Iinclude test.cpp
struct string
{
  constexpr string() = default;
  constexpr ~string() = default;

  gc_allocator<char> a;
};

int main()
{ }
```

This fails to compile before this commit and succeeds afterward. I am using these changes in practice within jank's [0] compiler and runtime here [1].

Note that `constexpr` itself was added in C++11, but it can't apply to destructors until C++20, and it's coming on 2024, so I've skipped over trying to juggle the different levels of `constexpr` here by just saying if you have C++20, you can do what we need.

[0]: https://github.com/jank-lang/jank
[1]: https://github.com/jank-lang/jank/pull/50/files#diff-e0a05b8607cd30211eacb4bc222986254ed8af3942ae602092ca3dbd150121d4

### DIFF
--- a/include/gc/gc_allocator.h
+++ b/include/gc/gc_allocator.h
@@ -138,34 +138,34 @@ public:
     typedef gc_allocator<GC_Tp1> other;
   };
 
-  gc_allocator() GC_NOEXCEPT {}
-  gc_allocator(const gc_allocator&) GC_NOEXCEPT {}
+  GC_ATTR_CONSTEXPR gc_allocator() GC_NOEXCEPT {}
+  GC_ATTR_CONSTEXPR gc_allocator(const gc_allocator&) GC_NOEXCEPT {}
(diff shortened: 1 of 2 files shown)


ivmai:
Looks good!

jeaye:
Nice, glad to hear it. Thanks for the quick review. Ready when you are.

ivmai:
Merged. Thank you!

Issue #602, opened 2023-12-25
Zig support improvement


ivmai:
Continuation of #557 and PR #598.

What to improve in build.zig:
- resolve TODO item: specify PACKAGE_VERSION and LIB*_VER_INFO, convert VER_INFO values to [SO]VERSION ones (moved to #638)
- [x] resolve TODO item: support enable_cplusplus
- [x] resolve TODO item: support enable_throw_bad_alloc_library
- [x] resolve TODO item: support build_cord
- [x] change test names to have "test" suffix as produced by cmake
- [x] support MacOS targets
- [x] remove -Dbuild_tests options

What to improve in GH Actions:
- [x] update to zig v0.12 final (when released)
- [x] support cross-compile for MacOS
- [x] build and test natively on MacOS
- [x] support cross-compile for other tier 1 targets
- [x] test with -Denable_cplusplus

ivmai:
@plajjan It would be very convenient to support CFLAGS_EXTRA option, could you please add it?
See CMakeLists.txt and Makefile.direct for the reference. (I'm not sure we need to separate the arguments in CFLAGS_EXTRA.)

plajjan:
With "replace -Denable_threads=false to -fsingle-threaded", do you mean we should replace the user visible input? I don't think that is possible. I think the idea with zig build is that you don't just let users pass compiler flags straight to the compiler but rather expose them as project specific options (the `-D` options) and then set compiler flags based on that. 

If on the other hand you mean that we should set `-fsingle-threaded` based on `-Denable_threads=false`, that should certainly be possible. Zig compilation objects, i.e. libraries and executable, have a single_threaded option on them, that we should probably set. I think that implies `-fsingle-threaded` down the line.

plajjan:
@ivmai yes, I can add a simple CFLAGS_EXTRA option that just passes through its value to the compiler. I don't see any added value in parsing it to separate arguments etc, so straight pass-through it is! :)

ivmai:
> With "replace -Denable_threads=false to -fsingle-threaded", do you mean we should replace the user visible input? I don't think that is possible. I think the idea with zig build is that you don't just let users pass compiler flags straight to the compiler but rather expose them as project specific options (the `-D` options) and then set compiler flags based on that.
> 
[...]

PR #609, opened 2024-01-10
Test zig build on MacOS too

plajjan:
This adds so the zig build test runs natively on MacOS too.

Part of #602.

### DIFF
--- a/.github/workflows/zig-build.yml
+++ b/.github/workflows/zig-build.yml
@@ -7,7 +7,7 @@ jobs:
   build:
     name: ${{ matrix.os }} thr:${{ matrix.enable_threads }} rwlock:${{ matrix.enable_rwlock }} redir:${{ matrix.redirect_malloc }} gcdeb:${{ matrix.enable_gc_debug }} munmap:${{ matrix.enable_munmap }} paramark:${{ matrix.parallel_mark }} thrlocal:${{ matrix.thread_local_alloc }}
     runs-on: ${{ matrix.os  }}
-    timeout-minutes: 4
+    timeout-minutes: 6
 
     strategy:
       fail-fast: false
--- a/.github/workflows/zig-build.yml
+++ b/.github/workflows/zig-build.yml
@@ -5,14 +5,15 @@ on: [push, pull_request]
 
 jobs:
   build:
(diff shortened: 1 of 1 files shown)


plajjan:
@ivmai looking at the workflow https://github.com/actonlang/bdwgc/actions/runs/7475855579 it seems like it fails with threads enabled but thread local allocation disabled. Not sure I'm reading it right though so probably better you take a look?

.github/workflows/zig-build.yml
```diff
@@ -39,9 +40,14 @@ jobs:
...
     - uses: actions/checkout@v4
-    - name: "Install zig"
+    - name: "Install zig on Linux"
+      if: runner.os == 'Linux'
       run: |
         mkdir zig && curl https://ziglang.org/builds/zig-linux-x86_64-0.12.0-dev.2076+8fd15c6ca.tar.xz | tar Jx --directory=zig --strip-components=1
+    - name: "Install zig on MacOS"
+      if: runner.os == 'macOS'
+      run: |
+        mkdir zig && curl https://ziglang.org/builds/zig-macos-x86_64-0.12.0-dev.2076+8fd15c6ca.tar.xz | tar Jx --directory=zig --strip-components=1
```
ivmai, line 50: Can we define ZIG_XZ_URL (probably in lower case) above and use it in 2 install lines?

plajjan: Yes but I picked this style because I reckoned it would better fit on windows where I imagine the commands are entirely different (I haven't really checked, just assume it's different).

ivmai: Aha, maybe not URL but just zig_version (to have a single place where ther version is specified per file)

ivmai:
> @ivmai looking at the workflow https://github.com/actonlang/bdwgc/actions/runs/7475855579 it seems like it fails with threads enabled but thread local allocation disabled. Not sure I'm reading it right though so probably better you take a look?

The code w/o thread-local allocation runs slower thus the test exceeds 4 minutes, pleasse change it to 6 (should be enough).

ivmai:
Are shared builds OK on MacOS?

.github/workflows/zig-build.yml
ivmai: > maybe not URL but just zig_version (to have a single place where ther version is specified per file)

I will change it myself

ivmai:
Merged as [9464650](https://github.com/ivmai/bdwgc/commit/94646503cceb0bfa9d4501b2ead7ace78a6712c2)

PR #618, opened 2024-02-04
Do not add EXTRA_BYTES twice in GC_generic_malloc_many()

### LINKED ISSUE #453: Performance degradation with multiple threads allocating
Hello,

We've been using the GC for several years now with great success. One issue we've come across though is a substantial performance degradation when multiple threads are calling `GC_malloc`, even with thread-local alloc enabled. In one application for example we have multiple threads creating very large linked lists in an embarrassingly parallel way, but we don't get any speedups with multithreading due to this issue.

Here is a simplified example showing the issue:

```c
#include <stdlib.h>
#include <stdio.h>
#include <time.h>
#include <omp.h>

#define GC_THREADS
#include <gc.h>

typedef struct list {
    long data1;
    long data2;
    long data3;
    struct list *next;
} list_t;

int main(void) {
    GC_INIT();
    GC_allow_register_threads();

    #pragma omp parallel
    {
        struct GC_stack_base sb;
        GC_get_stack_base(&sb);
        GC_register_my_thread(&sb);
    }

    struct timespec start, finish;
    clock_gettime(CLOCK_MONOTONIC, &start);

    #pragma omp parallel
    {
        list_t *curr = NULL;
        long data = 0;

        for (int i = 0; i < 100000000; i++) {
            list_t *node = GC_malloc(sizeof(*node));
            node->data1 = data + 1;
            node->data2 = data + 2;
            node->data3 = data + 3;
            ++data;
            if (curr)
                curr->next = node;
            curr = node;
        }
    }

    clock_gettime(CLOCK_MONOTONIC, &finish);
    double elapsed = (finish.tv_sec - start.tv_sec);
    elapsed += (finish.tv_nsec - start.tv_nsec) / 1000000000.0;
    printf("took %.2f seconds\n", elapsed);
}
```

We are using d0ba209660ea8c663e06d9a68332ba5f42da54ba -- on master this takes substantially longer and I haven't run it to completion. We build with CMake configured as:

```
cmake .. -DCMAKE_POSITION_INDEPENDENT_CODE=ON \
         -DBUILD_SHARED_LIBS=OFF \
         -Denable_threads=ON \
         -Denable_large_config=ON \
         -Denable_thread_local_alloc=ON \
         -Denable_handle_fork=ON
```

and link statically. Running this on an Apple M1 gives:

```
arshajii@arshajii-mbp misc % OMP_NUM_THREADS=1 ./test
took 5.55 seconds
arshajii@arshajii-mbp misc % OMP_NUM_THREADS=2 ./test
took 9.21 seconds
```

It seems like there is contention in `GC_malloc`. Any insight into why this happens and/or ways to mitigate it would be much appreciated.

As an aside I also tried a version that limits the linked list size to 100 and drops references to the old nodes. There is still a significant time increase on 2 threads: 2s vs 3.5s.

DaveBarton:
* mallocx.c (GC_generic_malloc_many): Document that EXTRA_BYTES has already been added to lb. Therefore subtract EXTRA_BYTES before calling GC_generic_malloc_inner(), which also adds it. Otherwise, especially when using thread-local storage, the heap can get filled up with objects of a slightly larger size than requested. When these are freed, they're put on their global free lists, instead of the ones for the actually requested sizes. Thus they're rarely found or transferred to the thread-local free lists of the main requested sizes by GC_generic_malloc_many(). Thread-local allocation is lost, and mutexes end up being raised on almost every allocation, which is very slow with multiple threads. (A similar slowdown happens if you explicitly call GC_FREE() too often from user code.)

This may also fix #453.

### DIFF
--- a/mallocx.c
+++ b/mallocx.c
@@ -252,7 +252,8 @@ GC_API size_t GC_CALL GC_get_expl_freed_bytes_since_gc(void)
 /* GC_malloc_many or friends to replenish it.  (We do not round up      */
 /* object sizes, since a call indicates the intention to consume many   */
 /* objects of exactly this size.)                                       */
-/* We assume that the size is a multiple of GC_GRANULE_BYTES.           */
+/* We assume that the size is a multiple of GC_GRANULE_BYTES, and that  */
+/* EXTRA_BYTES was already added to it.                                 */
 /* We return the free-list by assigning it to *result, since it is      */
 /* not safe to return, e.g. a linked list of pointer-free objects,      */
 /* since the collector would not retain the entire list if it were      */
@@ -433,7 +434,7 @@ GC_API void GC_CALL GC_generic_malloc_many(size_t lb, int k, void **result)
 
     /* As a last attempt, try allocating a single object.  Note that    */
     /* this may trigger a collection or expand the heap.                */
-      op = GC_generic_malloc_inner(lb, k, 0 /* flags */);
+      op = GC_generic_malloc_inner(lb - EXTRA_BYTES, k, 0 /* flags */);
       if (op != NULL) obj_link(op) = NULL;
 
   out:


ivmai:
Dave, nice catch!
I've changed your commit and merged. See 
9de8a9c. Please test it.

DaveBarton:
Thanks for renaming and improving/fixing the commit! We'll start using it now and we'll let you know if there are any problems.

mahrud:
@ivmai any chance you could make a new minor release with this change included?

ivmai:
Okay, in Jun.

PR #621, opened 2024-03-06
Provide mechanism to track allocations with heap profiling tools

jacob-hughes:
Tools such as Valgrind's massif[1] and KDE heaptrack[2] work by intercepting dynamically linked allocator calls so that they can do their own bookkeeping for tracking memory usage. While this enables them to trivially hook into any of the BDWGC calls which allocate, they aren't currently able to tell when the GC frees memory.

This commit introduces a new function to the API, `GC_free_profiler_hook`, which is called for each object during the reclaim phase of the GC. It provides a stub for heap profiling tools to override so that they can see which objects were freed.

This functionality is feature gated behind the `-DVALGRIND_TRACKING` flag.

[1]: https://valgrind.org/docs/manual/ms-manual.html
[2]: https://github.com/KDE/heaptrack

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -87,6 +87,7 @@ option(enable_disclaim "Support alternative finalization interface" ON)
 option(enable_dynamic_pointer_mask "Support pointer mask/shift set at runtime" OFF)
 option(enable_large_config "Optimize for large heap or root set" OFF)
 option(enable_gc_assertions "Enable collector-internal assertion checking" OFF)
+option(enable_track_valgrind "Support tracking GC_malloc and friends for heap profiling tools" OFF)
 option(enable_mmap "Use mmap instead of sbrk to expand the heap" OFF)
 option(enable_munmap "Return page to the OS if empty for N collections" ON)
 option(enable_dynamic_loading "Enable tracing of dynamic library data roots" ON)
@@ -101,7 +102,6 @@ option(enable_emscripten_asyncify "Use Emscripten asyncify feature" OFF)
 option(install_headers "Install header and pkg-config metadata files" ON)
 option(with_libatomic_ops "Use an external libatomic_ops" OFF)
 option(without_libatomic_ops "Use atomic_ops.h in libatomic_ops/src" OFF)
-option(enable_track_valgrind "Support tracking GC_malloc and friends for heap profiling tools" OFF)
 
 # Override the default build type to RelWithDebInfo (this instructs cmake to
 # pass -O2 -g -DNDEBUG options to the compiler by default).
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -101,6 +101,7 @@ option(enable_emscripten_asyncify "Use Emscripten asyncify feature" OFF)
 option(install_headers "Install header and pkg-config metadata files" ON)
 option(with_libatomic_ops "Use an external libatomic_ops" OFF)
 option(without_libatomic_ops "Use atomic_ops.h in libatomic_ops/src" OFF)
+option(enable_track_valgrind "Support tracking GC_malloc and friends for heap profiling tools" OFF)
 
 # Override the default build type to RelWithDebInfo (this instructs cmake to
 # pass -O2 -g -DNDEBUG options to the compiler by default).
@@ -713,6 +714,10 @@ if (build_cord)
           RUNTIME DESTINATION "${CMAKE_INSTALL_BINDIR}"
           INCLUDES DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}")
 endif(build_cord)
+
+if (enable_valgrind_tracking)
+  add_definitions("-DVALGRIND_TRACKING")
+endif()
 
 if (BUILD_SHARED_LIBS AND HAVE_FLAG_WL_NO_UNDEFINED)
   # Declare that the libraries do not refer to external symbols.
--- a/allchblk.c
+++ b/allchblk.c
@@ -1005,7 +1005,7 @@ STATIC struct hblk *GC_allochblk_nth(size_t lb_adjusted, int k,
     return hbp;
 }
 
-# if defined(VALGRIND_TRACKING)
+# ifdef VALGRIND_TRACKING
     GC_API void GC_CALL GC_free_profiler_hook(void * p) {
         UNUSED_ARG(p);
     }
--- a/allchblk.c
+++ b/allchblk.c
@@ -1008,7 +1008,7 @@ STATIC struct hblk *GC_allochblk_nth(size_t lb_adjusted, int k,
 # ifdef VALGRIND_TRACKING
 /* This function is called from the sweeper whenever an object is      */
 /* freed.                                                              */
-    GC_API void GC_CALL GC_free_profiler_hook(void * p) {
+    GC_ATTR_NOINLINE GC_API void GC_CALL GC_free_profiler_hook(void * p) {
         GC_noop1((word)(p));
     }
 # endif
--- a/allchblk.c
+++ b/allchblk.c
@@ -1005,6 +1005,12 @@ STATIC struct hblk *GC_allochblk_nth(size_t lb_adjusted, int k,
     return hbp;
 }
 
+# if defined(VALGRIND_TRACKING)
+    GC_API void GC_CALL GC_free_profiler_hook(void * p) {
+        UNUSED_ARG(p);
+    }
+# endif
+
 /*
  * Free a heap block.
  *
--- a/allchblk.c
+++ b/allchblk.c
@@ -1006,6 +1006,8 @@ STATIC struct hblk *GC_allochblk_nth(size_t lb_adjusted, int k,
 }
 
 # ifdef VALGRIND_TRACKING
+/* This function is called from the sweeper whenever an object is      */
+/* freed.                                                              */
     GC_API void GC_CALL GC_free_profiler_hook(void * p) {
         GC_noop1((word)(p));
     }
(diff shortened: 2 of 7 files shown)


include/gc/gc.h
```diff
@@ -605,6 +605,20 @@ GC_API int GC_CALL GC_posix_memalign(void ** /* memptr */, size_t /* align */,
 /* GC_free(0) is a no-op, as required by ANSI C for free.               */
 GC_API void GC_CALL GC_free(void *);
 
+# if defined(VALGRIND_TRACKING)
```
ivmai, line 608: By convention, ifdef is not use in public header, where possible (except for ifdef GC_THREADS), please remove it. See the next comment.

include/gc/gc.h
```diff
@@ -605,6 +605,20 @@ GC_API int GC_CALL GC_posix_memalign(void ** /* memptr */, size_t /* align */,
...
+/* Notify heap profiling tools that an object was deallocated.          */
+/* Programs such as Valgrind massif and KDE heaptrack track allocated   */
+/* objects by overriding common allocator methods (e.g. malloc and      */
+/* free). However, because GC doesn't work by calling a standard alloc  */
+/* methods on objects which were reclaimed, we need a way to tell       */
+/* profilers that an object has been freed.                             */
+/*                                                                      */
+/* This function is called from the sweeper whenever an object is       */
+/* freed, which can then be intercepted by heap profilers so that       */
+/* they can accurately track allocations.                               */
```
ivmai, line 618: Add comment: Defined only if the library has been compiled with VALGRIND_TRACKING.

CMakeLists.txt
```diff
@@ -101,6 +101,7 @@ option(enable_emscripten_asyncify "Use Emscripten asyncify feature" OFF)
 option(install_headers "Install header and pkg-config metadata files" ON)
 option(with_libatomic_ops "Use an external libatomic_ops" OFF)
 option(without_libatomic_ops "Use atomic_ops.h in libatomic_ops/src" OFF)
+option(enable_track_valgrind "Support tracking GC_malloc and friends for heap profiling tools" OFF)
```
ivmai, line 104: Please also add support to build.zig

configure.ac
```diff
@@ -1032,6 +1032,14 @@ if test "${enable_gc_assertions}" = yes; then
               [Define to enable internal debug assertions.])
 fi
 
+AC_ARG_ENABLE(valgrind-tracking,
+    [AS_HELP_STRING([--enable-valgrind-tracking],
+        [heap profiler allocation tracking])])
+if test "${enable_valgrind_tracking}" = yes; then
+    AC_DEFINE([VALGRIND_TRACKING], 1,
+              [Define to enable heap profiler allocation tracking.])
```
ivmai, line 1040: I think it is better to have the message matching that in cmake script, e.g.: Define to support tracking ...

CMakeLists.txt
ivmai, line 104: I suggest to move this line upper to follow option enable_gc_assertions (to have similar order of options as in configure)

allchblk.c
```diff
@@ -1005,6 +1005,12 @@ STATIC struct hblk *GC_allochblk_nth(size_t lb_adjusted, int k,
     return hbp;
 }
 
+# if defined(VALGRIND_TRACKING)
```
ivmai, line 1008: `#ifdef VALGRIND_TRACKING` (fix this everywhere)

allchblk.c
```diff
@@ -1005,6 +1005,12 @@ STATIC struct hblk *GC_allochblk_nth(size_t lb_adjusted, int k,
...
+    GC_API void GC_CALL GC_free_profiler_hook(void * p) {
+        UNUSED_ARG(p);
```
ivmai, line 1010: I guess we should use GC_noop1((word)p) instead. Otherwise a compiler with -O3 option could remove such calls from the code.

reclaim.c
```diff
@@ -200,6 +200,10 @@ STATIC ptr_t GC_reclaim_clear(struct hblk *hbp, hdr *hhdr, word sz,
                 obj_link(p) = list;
                 list = p;
 
+#             if defined(VALGRIND_TRACKING)
+                GC_free_profiler_hook(p);
+#             endif
+
```
ivmai, line 206: Del blank line here

include/gc/gc.h
```diff
@@ -605,6 +605,20 @@ GC_API int GC_CALL GC_posix_memalign(void ** /* memptr */, size_t /* align */,
 /* GC_free(0) is a no-op, as required by ANSI C for free.               */
 GC_API void GC_CALL GC_free(void *);
 
+# if defined(VALGRIND_TRACKING)
+/* Notify heap profiling tools that an object was deallocated.          */
```
ivmai, line 609: Notify heap profiling tools that an object was deallocated.
->
A symbol to be intercepted by heap profilers so that they can accurately track allocations.

allchblk.c
ivmai, line 1008: Add comment: This function is called from the sweeper whenever an object is freed.

ivmai, line 1008: Add: GC_ATTR_NOINLINE
I'm not sure about the order: GC_ATTR_NOINLINE GC_API or the opposite (should be checked with popular compilers - clang, gcc, msvc cl)

include/gc/gc.h
```diff
@@ -605,6 +605,20 @@ GC_API int GC_CALL GC_posix_memalign(void ** /* memptr */, size_t /* align */,
...
+/* Programs such as Valgrind massif and KDE heaptrack track allocated   */
+/* objects by overriding common allocator methods (e.g. malloc and      */
+/* free). However, because GC doesn't work by calling a standard alloc  */
+/* methods on objects which were reclaimed, we need a way to tell       */
+/* profilers that an object has been freed.                             */
+/*                                                                      */
```
ivmai, line 615: This function is called from the sweeper whenever an object is freed, which can then be intercepted by heap profilers so that they can accurately track allocations.
->
This function is not intended to be called by the client, it should be used for the interception purpose only.

ivmai (changes requested):

allchblk.c
ivmai: But probably it is ok to leave UNUSED if no in line is added.

include/gc/gc.h
jacob-hughes: Fixed.

include/gc/gc.h
jacob-hughes: Fixed.

CMakeLists.txt
jacob-hughes: Fixed.

configure.ac
jacob-hughes: Good point, fixed.

CMakeLists.txt
jacob-hughes: Sure, fixed.

allchblk.c
jacob-hughes: Fixed

allchblk.c
jacob-hughes: That makes sense, I've fixed this now :)

include/gc/gc.h
jacob-hughes: Fixed

allchblk.c
jacob-hughes: Fixed

jacob-hughes: Fixed

include/gc/gc.h
jacob-hughes: Fixed

malloc.c
```diff
@@ -651,6 +651,10 @@ GC_API void GC_CALL GC_free(void * p)
         return;
     }
 
+#   ifdef VALGRIND_TRACKING
+      GC_free_profiler_hook(p);
+#   endif
```
ivmai, line 656: del blank line here

malloc.c
```diff
@@ -651,6 +651,10 @@ GC_API void GC_CALL GC_free(void * p)
         return;
     }
 
+#   ifdef VALGRIND_TRACKING
```
ivmai, line 654: I think it would be better to remove "ifdef VALGRIND_TRACKING" everywhere but defining GC_free_profiler_hook(p) macro as (void)(p) in gc_priv.h if VALGRIND_TRACKING is not defined.

allchblk.c
```diff
@@ -1005,6 +1005,14 @@ STATIC struct hblk *GC_allochblk_nth(size_t lb_adjusted, int k,
     return hbp;
 }
 
+# ifdef VALGRIND_TRACKING
+/* This function is called from the sweeper whenever an object is      */
+/* freed.                                                              */
+    GC_ATTR_NOINLINE GC_API void GC_CALL GC_free_profiler_hook(void * p) {
+        GC_noop1((word)(p));
```
ivmai, line 1012: (p) -> p (i.e. del extra parens)

ivmai (changes requested):
Some cosmetic changes are still needed

ivmai:
I will squash and polish the code myself. Will do it today or tomorrow.

malloc.c
jacob-hughes: I've probably misunderstood this. Do you want me to replace the `GC_free_profiler_hook` function with a macro definition? I don't think this would work anymore as there would be no symbol for profilers to intercept.

malloc.c
jacob-hughes: Fixed

jacob-hughes:
Thanks @ivmai, apologies for the delay on this one.

malloc.c
ivmai: No, I will do it, in some other way.

ivmai:
Squashed, modified and committed as [a73e179](https://github.com/ivmai/bdwgc/commit/a73e17946ff8d3a865077489ed8d9a41c211ac36).
Please test it.

PR #625, opened 2024-03-18
build.zig: Support ignore_free

plajjan:
Add new enable_ignore_free option to build.zig and fix an actual bug related to the implementation of ignore_free. That it was a comment must have been a mistake. I made the original comment that adds the `UNUSED_ARG(p)` line so no idea how I messed that up, perhaps some rebase issue, but this fixes it!

### DIFF
--- a/malloc.c
+++ b/malloc.c
@@ -698,7 +698,7 @@ GC_API void GC_CALL GC_free(void * p)
   void free(void * p)
   {
 #   ifdef IGNORE_FREE
-#     UNUSED_ARG(p);
+      UNUSED_ARG(p);
 #   else
 #     if defined(GC_LINUX_THREADS) && !defined(USE_PROC_FOR_LIBRARIES)
         /* Don't bother with initialization checks.  If nothing         */
--- a/build.zig
+++ b/build.zig
@@ -88,6 +88,8 @@ pub fn build(b: *std.Build) void {
         "Support for atomic uncollectible allocation") orelse true;
     const enable_redirect_malloc = b.option(bool, "enable_redirect_malloc",
         "Redirect malloc and friend to GC routines") orelse false;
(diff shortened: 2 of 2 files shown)


malloc.c
```diff
@@ -698,7 +698,7 @@ GC_API void GC_CALL GC_free(void * p)
   void free(void * p)
   {
 #   ifdef IGNORE_FREE
-#     UNUSED_ARG(p);
```
ivmai, line 701: Good catch! I will add testing of it as well.

plajjan:
Okay. Might I ask about the reasoning?

It seems to me that I need ignore free when used with redirect malloc and other people's libraries that do explicit frees, no?

ivmai:
> It seems to me that I need ignore free when used with redirect malloc

Why? What's the purpose?
Generally malloc/free/realloc redirection is used for 2 purposes:
- leak detection
- force 3rd-party code e.g. containers (which uses malloc) to allocate object from the GC heap so that to have such objects traceable by GC (i.e. able to store objects allocated by GC_malloc to a container which implementation uses plain malloc).

ivmai:
> fix an actual bug related to the implementation of ignore_free
Done by commit d76963c2bb3.
Thank you!

plajjan:
Hmm ok. I started noticing problems when I implemented fs related functions in Acton. Previously all test programs had been running fine but doing uv_fs_scandir() would segfault the program. I don't have the backtraces in front of me right now (but should be able to reproduce) but something led me to believe to a double-free-looking scenario or perhaps it's running GC_free on something allocated via normal malloc. Enabling IGNORE_FREE fixed the problem so the segfault went away.

I sort of just presumed actually that the preferred option when doing link time redircetion would be to just ignore calls to free() and instead let the GC collect all unreachable objects. Why do we actually want to support explicit free() at all?

Issue #396, opened 2021-12-01
Non-zero suspend_ack_sem assertion failure in suspend_restart_barrier


ivmai:
Source: master (fd8e8cb)
Build: https://app.travis-ci.com/github/ivmai/bdwgc/jobs/550328480
Host: Linux/x64
Config: CFLAGS_EXTRA="-fsanitize=thread -D NO_CANCEL_SAFE -D NO_INCREMENTAL -fno-omit-frame-pointer" CONF_OPTIONS="--disable-parallel-mark --disable-thread-local-alloc --disable-shared --enable-gc-assertions --without-libatomic-ops"

ivmai:
Build: https://app.travis-ci.com/github/ivmai/bdwgc/jobs/563521118
Source: release-8_2 (2d93163e8a8)
Config:
* CFLAGS_EXTRA="-fsanitize=thread -D NO_CANCEL_SAFE -D NO_INCREMENTAL -D USE_SPIN_LOCK -fno-omit-frame-pointer -D TEST_FORK_WITHOUT_ATFORK" 
* CONF_OPTIONS="--enable-gc-assertions --enable-gc-debug --enable-handle-fork=manual --enable-large-config --with-libatomic-ops=no"

Output:
Assertion failure: extra/../pthread_stop_world.c:465

Code:
```
      sem_getvalue(&GC_suspend_ack_sem, &i);
      GC_ASSERT(0 == i);
```

PR #633, opened 2024-04-11
Fix dismatch between GC_suspend_ack_sem and n_live_threads.

delguoqing:
### The cause of this problem

This is a corner case I found in Android, but this also applies for other platforms, as long as they have chances of receiving signals and therefore pausing their threads.

When the user or system want to create a bugreport, it will send a SIGQUIT to each process. Because android app is forked from zygote, they all have a Signal Catcher thread that handles SIGQUIT. The signal catcher thread will in turn send SIGRT_1 to all of the other threads to pause them and and get the ucontext, then do a stack unwind.

1.When the GC_stopping_thread have finished collecting and want to restart the world, it first change GC_stop_count to GC_stop_count + 1. 

2.Just before the GC_stopping_thread have a chance to raise signals to other threads by calling GC_restart_all. Other threads might have been awaken by other signals than GC_sig_thr_restart(in my case SIGRT_1), then because of the updated value of GC_stop_count. The following loop condition will not meet.

```
  do {
      sigsuspend(&suspend_handler_mask);
      /* Iterate while not restarting the world or thread is suspended. */
  } while (ao_load_acquire_async(&GC_stop_count) == my_stop_count
#          ifdef GC_ENABLE_SUSPEND_THREAD
             || ((suspend_cnt & 1) != 0
                 && (word)ao_load_async(&(me -> ext_suspend_cnt))
                    == suspend_cnt)
#          endif
          );

```
The threads can go straight to the end of the suspend_handler and update last_stop_count.
```
  {
    /* If the RESTART signal loss is possible (though it should be      */
    /* less likely than losing the SUSPEND signal as we do not do       */
    /* much between the first sem_post and sigsuspend calls), more      */
    /* handshaking is provided to work around it.                       */
    sem_post(&GC_suspend_ack_sem);
    /* Set the flag that the thread has been restarted. */
    if (GC_retry_signals)
      ao_store_release_async(&(me -> last_stop_count),
                             my_stop_count | THREAD_RESTARTED);
  }
  RESTORE_CANCEL(cancel_state);
```

3. The GC_stopping_thread then continues and call GC_restart_all, in order to send GC_sig_thr_restart to all other threads. However, it sees that some threads have their last_stop_count already updated and decide that it will not raise signal to it "again". So n_live_threads is not counting those threads that runs too fast in step 2.

```
    for (i = 0; i < THREAD_TABLE_SZ; i++) {
      for (p = GC_threads[i]; p != NULL; p = p -> tm.next) {
        if (!THREAD_EQUAL(p -> id, self)) {
          if ((p -> flags & (FINISHED | DO_BLOCKING)) != 0) continue;
#         ifdef GC_ENABLE_SUSPEND_THREAD
              if ((p -> ext_suspend_cnt & 1) != 0) continue;
#         endif
          if (GC_retry_signals
                && AO_load(&(p -> last_stop_count)) == GC_stop_count)
              continue; /* The thread has been restarted. */
          n_live_threads++;
#         ifdef DEBUG_THREADS
            GC_log_printf("Sending restart signal to %p\n", (void *)p->id);
#         endif
```

4. In the end, we can have n_live_threads less than GC_suspend_ack_sem. So we will call sem_wait only n_live_threads times, thus leaving GC_suspend_ack_sem a non-zero value after this round of GC.

### My fix for this problem

I suggest that we raise signal to all of the threads for the first time, if we're not actually retrying.
I added a boolean argument for GC_suspend_all/GC_restart_all to indicate if we're actually doing the retry. If not, we ignore last_stop_count and always raise signals to suspending threads.

### DIFF
--- a/pthread_stop_world.c
+++ b/pthread_stop_world.c
@@ -438,7 +438,7 @@ static void suspend_restart_barrier(int n_live_threads)
 # define WAIT_UNIT 3000 /* us */
 
 static int resend_lost_signals(int n_live_threads,
-                               int (*suspend_restart_all)(void))
+                               int (*suspend_restart_all)(GC_bool))
 {
 #   define RETRY_INTERVAL 100000 /* us */
 #   define RESEND_SIGNALS_LIMIT 150
@@ -455,7 +455,7 @@ static int resend_lost_signals(int n_live_threads,
         if (ack_count == n_live_threads)
           break;
         if (wait_usecs > RETRY_INTERVAL) {
-          int newly_sent = suspend_restart_all();
+          int newly_sent = suspend_restart_all(GC_retry_signals);
 
           if (newly_sent != prev_sent) {
             retry = 0; /* restart the counter */
@@ -488,7 +488,7 @@ static int resend_lost_signals(int n_live_threads,
 #endif
 
 static void resend_lost_signals_retry(int n_live_threads,
-                                      int (*suspend_restart_all)(void))
+                                      int (*suspend_restart_all)(GC_bool))
 {
 # if defined(HAVE_CLOCK_GETTIME) && !defined(DONT_TIMEDWAIT_ACK_SEM)
 #   define TIMEOUT_BEFORE_RESEND 10000 /* us */
@@ -912,7 +912,7 @@ GC_INNER void GC_push_all_stacks(void)
 
 /* Suspend all threads that might still be running.  Return the number  */
 /* of suspend signals that were sent.                                   */
-STATIC int GC_suspend_all(void)
+STATIC int GC_suspend_all(GC_bool is_retry)
 {
   int n_live_threads = 0;
   int i;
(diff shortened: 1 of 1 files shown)


ivmai:
Probably related: #396

ivmai:
> I suggest that we raise signal to all of the threads for the first time, if we're not actually retrying.

I've switched to this problem, let me think how to do handle spurious wakeups properly.

ivmai:
> I suggest that we raise signal to all of the threads for the first time, if we're not actually retrying.

I have significantly reworked your commit (by introducing a static variable instead of an argument) but the high-level logic remains the same. Please test it.

ivmai:
Also, I've added a FIXME because there still could be a chance of a mismatch of ack_sem vs n_live_threads when we are retrying (i.e. a signal is lost and sem wait timed out and then the thread is restarted by SIGQUIT), but the chance of this is much smaller compared to this rare case.

PR #643, opened 2024-05-13
Add skip_cpp_definitions to gc_cpp.cc

fruffy:
Trying to compile a static BDWGC build with your own overrides for the allocation functions can lead to a duplicate definition error for the C++ operators (new, delete, new[],..). This PR adds a define directive to skip these definitions, if requested.

### DIFF
--- a/docs/README.macros
+++ b/docs/README.macros
@@ -269,7 +269,7 @@ REDIRECT_FREE=<X>       Causes free to be redirected to X.  The canonical use
   is REDIRECT_FREE=GC_debug_free.
 
 SKIP_CPP_DEFINITIONS    Do not include definitions for new, delete, etc in 
-  gc_cpp.h. This might be necessary for static compilation for user-supplied 
+  gc_cpp.cc. This might be necessary for static compilation for user-supplied 
   allocation definitions.
 
 IGNORE_FREE     Turns calls to free into a no-op.  Only useful with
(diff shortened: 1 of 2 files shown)


ivmai:
Okay

docs/README.macros
```diff
@@ -268,6 +268,10 @@ REDIRECT_REALLOC=<X>    Causes realloc to be redirected to X.
 REDIRECT_FREE=<X>       Causes free to be redirected to X.  The canonical use
   is REDIRECT_FREE=GC_debug_free.
 
+SKIP_CPP_DEFINITIONS    Do not include definitions for new, delete, etc in 
+  gc_cpp.h. This might be necessary for static compilation for user-supplied
```
fruffy, line 272:
```suggestion
  gc_cpp.cc. This might be necessary for static compilation for user-supplied 
```

ivmai:
Is it OK if I change your email of the squashed commit to contact at ruffy.eu ?

fruffy:
Sure!

ivmai:
Hmm, I don't quite understand the issue: if you don't need the definitions in gc_cpp.cc then why do you link with libgccpp.a (or .so)?

ivmai:
Anyway, I've merged the patch but changing macro name from SKIP_CPP_DEFINITIONS to SKIP_GCCPP_DEFINITIONS.

fruffy:
Thank you! 

> Hmm, I don't quite understand the issue: if you don't need the definitions in gc_cpp.cc then why do you link with libgccpp.a (or .so)?

We also need libgccpp.a because of `GC_throw_bad_alloc` but since it just one file we could also directly compile it..

ivmai:
> We also need libgccpp.a because of GC_throw_bad_alloc

Use libgctba.a instead

fruffy:
> > We also need libgccpp.a because of GC_throw_bad_alloc
> 
> Use libgctba.a instead

Must have missed this one, let me try.

Issue #646, opened 2024-06-04
Compilation with TCC fails on ARM macs


medvednikov:
```
tcc -c -I include/ gc.c
gc.c:14208: error: ARM asm not implemented.
```

We use an amalgamation of bdwgc:

https://github.com/vlang/v/blob/master/thirdparty/libgc/gc.c

Would it be be possible to replace asm with C inside an `#ifdef __TINYC__`?

Compilation on Linux and Intel macs works, this is a pure ARM mac + tcc issue.

Thanks

ivmai:
It is OK to use GC_noop1 instead, like on Intel compiler. Please make a PR

medvednikov:
Thanks! That indeed worked. Will submit a PR today.

medvednikov:
https://github.com/ivmai/bdwgc/pull/647

PR #647, opened 2024-06-07
Make compile on ARM macs with tinyc

medvednikov:
As discussed in https://github.com/ivmai/bdwgc/issues/646

### DIFF
--- a/include/private/gc_priv.h
+++ b/include/private/gc_priv.h
@@ -2632,7 +2632,12 @@ GC_EXTERN GC_bool GC_print_back_height;
 # define GC_auto_incremental (GC_incremental && !GC_manual_vdb)
   GC_INNER void GC_dirty_inner(const void *p); /* does not require locking */
 # define GC_dirty(p) (GC_manual_vdb ? GC_dirty_inner(p) : (void)0)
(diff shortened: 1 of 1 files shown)


ivmai:
Sorry for a delay.
Which tcc version do you use?

ivmai:
Also, I think it would be better to make a workaround to GC_reachable_here() definition, not REACHABLE_AFTER_DIRTY(). (I can change it myself.)

medvednikov:
We use the latest build from their `mob` branch. The don't do versioned releases often.

> Also, I think it would be better to make a workaround to GC_reachable_here() definition, not REACHABLE_AFTER_DIRTY()
Yes, that's a better place. Thanks.

ivmai:
> Which tcc version do you use?

I mean I'd like to put tcc version in the comment near the workaround.
I guess you use tcc-0.9.27, right? (https://download.savannah.nongnu.org/releases/tinycc/)

medvednikov:
```
tcc -v
tcc version 0.9.28rc
```

ivmai:
OK (https://github.com/TinyCC/tinycc/blob/mob/VERSION)

ivmai:
Merged with some modifications. Please test.

PR #649, opened 2024-08-01
use g instead of X for tinyc asm constraint

kimshrier:
The Tiny C compiler does not handle the `X` asm constraint.  Use the `g`  constraint instead.

### DIFF
--- a/include/gc/gc.h
+++ b/include/gc/gc.h
@@ -1537,6 +1537,9 @@ GC_API int GC_CALL GC_invoke_finalizers(void);
 # if defined(__e2k__)
 #   define GC_reachable_here(ptr) \
                 __asm__ __volatile__ (" " : : "r"(ptr) : "memory")
+# elif defined(__TINYC__)
+#   define GC_reachable_here(ptr) \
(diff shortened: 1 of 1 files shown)


ivmai:
It seems to supercede PR #647 . Could we remove a TinyCC macro check above?
Also, please tell us which TinyCC version do you use.

ivmai:
My question is: should we remove these 2 lines after applying your patch?
```
&& !(defined(__APPLE__) && defined(__arm__) && defined(__TINYC__))
        /* TCC (as of v0.9.28rc) does not support asm on macOS/arm. */
```

ivmai:
Processing the patch...

> The Tiny C compiler does not handle the X asm constraint.

Would be nice to have a link here to the corresponding issue in the TinyCC community, if any.

ivmai:
> The Tiny C compiler does not handle the X asm constraint. Use the g constraint instead.

Merged.

> I think the test in the 2 lines you quote above, is too specific. Tiny C does not support asm on arm for any OS.

Let's wait for reply from  @medvednikov and do it in a separate PR.

PR #651, opened 2024-08-01
Update README.emscripten

### LINKED ISSUE #650: Emscripten support fundamentally broken?
### Summary

I recently encountered an issue with bdwgc in emscripten where pointers "on the stack" (reason for quotation marks is explained below) are not found during the marking phase and thus memory is released too early. I believe that this is a not a simple bug but rather a fundamental issue that cannot be resolved without cooperation from the emscripten C compiler.

### A test case

A simple test program which shows the issue is as follows:

```C
#include <stdio.h>
#include "gc/gc.h"

struct list {
   int value;
   struct list *rest;
};

void print_list(struct list *p) {
   while (p->rest != NULL) {
      printf("%d ", p->value);
      p = p->rest;
   }
   printf("\n");
}

int main() {
   printf("Starting to allocate\n");
   struct list *head = GC_MALLOC(sizeof(struct list));
   struct list *p = head;
   head->value = 0;
   for (int i = 1; i < 10000; i++) {
      struct list *q = GC_MALLOC(sizeof(struct list));
      q->value = i;
      p->rest = q;
      p = q;
   }
   p->rest = NULL;
   print_list(head);
   return 0;
}
```

If everything works correctly, this should just print the numbers from 0 to 9999. But when compiled with `emcc -O2 test.c libgc.a -o test.html -sASYNCIFY` and run in node or in a web browser, it only prints numbers from 8192 upwards. The reason for the failure is that bdwgc does not find the `head` pointer during the marking phase and thus frees the entire list too early, reusing the  memory for later allocations.

### Why I think this issue is unsolvable without changes in the emscripten C compiler

Unlike ordinary assembly languages, webassembly does not have a fixed register file. Instead, there is the concept of a "local". These locals effectively behave as an unlimited number of registers which are automatically saved and restored before/after function calls by the wasm virtual machine. The `head` pointer in the test case is stored in such a local variable. These are not scanned by bdwgc and thus memory is reclaimed too early. However, to the best of my knowledge, there is no way to access the value of locals for functions further up the call stack. They are not saved on the same stack that emscripten uses and which bdwgc does scan for pointers, but rather in a memory space that is internal to the wasm VM.

Therefore, the issue does not seem to be solvable without instructing the C compiler to always keep pointers on a stack which is accessible to bdwgc. This usually happens when compiling with `-O0` (the above test case works correctly in this case), but I have also encountered examples where even with `-O0` pointers are not kept on the emscripten stack.

Unless there is some way around the problem that I did not think of, in my opinion it would be useful to warn about this issue in the documentation, if not remove the broken emscripten support altogether.

spaghettisalat:
Closes #650

### DIFF
--- a/docs/platforms/README.emscripten
+++ b/docs/platforms/README.emscripten
@@ -2,6 +2,25 @@
 The build system (at least autotools-based) should detect and configure
 emscripten correctly.
 
+Due to limitations of webassembly, finding all roots on the stack
+needs more coordination with the code using the gc than on other
+platforms. There are two strategies for dealing with this issue:
+
+1. Compile any code using the gc with
+   -sBINARYEN_EXTRA_PASSES='--spill-pointers'. This instructs the C
+   compiler to always keep roots in a place where the gc can find them.
+   This is the simplest option but there is some impact on code size and
+   run time.
+
+2. Only trigger the gc at points where it can be guaranteed that there
+   are no pointers on the stack. When running in the browser, this can
+   be done for instance in the main event loop using
+   emscripten_set_timeout(). Triggering the gc manually involves
+   calling GC_enable(), GC_collect() and GC_disable() in succession.
+   This method does not impact code size or runtime but might lead to
+   the gc running too often or not often enough if the timeouts are
+   chosen incorrectly.
+
 As of now, gctest almost passes, except for the tests that involve a_get().
 
 No thread support for now. No idea how to stop other threads (perhaps we need
@@ -10,6 +29,6 @@ support from JS side).
 How to build:
 
     # source EMSDK first
-    emconfigure ./configure
+    LDFLAGS="-sBINARYEN_EXTRA_PASSES='--spill-pointers'" emconfigure ./configure
     emmake make gctest.html
-    # point your browser at .libs/gctest.html
+    # point your browser at .libs/gctest.html or call `node .libs/gctest.js`


ivmai:
@jul Do you agree with the readme change?

jul:
wrong @jul « same player shoot again »©

juj:
I would recommend expanding on the comment a little bit.. right now it might give a misimpression that adding the spill-pointers pass or scanning the stack would be unconditionally required - which it isn't, if user only triggers a GC at points when it is known there are no managed objects on the stack.

So I would recommend stating the readme in a way that would illustrate that there are these two strategies that one can use, and also illustrate about the drawbacks of each. The challenge with using --spill-pointers is that it greatly increases generated code size on disk, and there may also be a small % performance penalty to doing so.

spaghettisalat:
Do you have some example code for the strategy where one triggers the GC only at specific points which we can link to in the README? I guess this requires some modifications of the bdwgc source code to avoid the automatic GC runs, right?

At the moment, your description of how this strategy works is somewhat cryptic to me and therefore it is hard for me to write a good summary for the README.

ivmai:
GC_disable() might be useful to avoid automatic GC

juj:
Yeah, `GC_disable()`, and then one can trigger an asynchronous GC e.g. by using [emscripten_set_timeout()](https://github.com/emscripten-core/emscripten/blob/main/system/include/emscripten/eventloop.h#L18) function that calls to a callback that manually runs `GC_collect();` (I forget if bdwgc requires one to call GC_enable() before calling GC_collect() manually?)

ivmai:
> I forget if bdwgc requires one to call GC_enable() before calling GC_collect() manually?

Yes, GC_enable() is needed otherwise GC_collect() is no-op.

spaghettisalat:
Alright, I have extended the description in the README file a bit. I hope that should do it.

juj:
Looks really good!

ivmai:
Merged.

PR #655, opened 2024-08-23
Move atomic_ops libs into private so that it won't confuse downstream packages

fundawang:
Acturally, bdwgc it self may requires atomic_ops, but it does not require downstream packages (like guile) requires atomic_ops. So I would suggest that move `@ATOMIC_OPS_LIBS@` into `Libs.private`, so that won't confuse downstream packages, especially when performing check.

### DIFF
--- a/bdw-gc.pc.in
+++ b/bdw-gc.pc.in
@@ -6,5 +6,6 @@ includedir=@includedir@
 Name: Boehm-Demers-Weiser Conservative Garbage Collector
 Description: A garbage collector for C and C++
 Version: @PACKAGE_VERSION@
-Libs: -L${libdir} @ATOMIC_OPS_LIBS@ -lgc @THREADDLLIBS@
+Libs: -L${libdir} -lgc @THREADDLLIBS@
+Libs.private: @ATOMIC_OPS_LIBS@
 Cflags: -I${includedir}


ivmai:
> Acturally, bdwgc it self may requires atomic_ops, but it does not require downstream packages (like guile) requires atomic_ops

This is not full true (see gc_inline.h) but most clients do not use that header. So, I think your patch worth to be applied.

ivmai:
We do not need to change CMakeLists.txt similarly, right?

ivmai:
One more question: does autoconf v2.61 support Libs.private? Otherwise we need to update AC_PREREQ(2.61)

fundawang:
> We do not need to change CMakeLists.txt similarly, right?
Yes, CMakeLists.txt does not need to be changed.

> One more question: does autoconf v2.61 support Libs.private? Otherwise we need to update AC_PREREQ(2.61)
configure.ac says `AC_CONFIG_FILES([Makefile bdw-gc.pc])`. It does not matter what `bdw-gc.pc.in` contains, it just replace the corresponding `@ATOMIC_OPS_LIBS@` variables. So basically, it does not affect the autoconf version the project requries.

ivmai:
> It does not matter what bdw-gc.pc.in contains

Okay, my question was incorrect. I'd better ask about pkg-config. If libs.private has been supported in pkg-config for ages, then it is fine.

fundawang:
> > It does not matter what bdw-gc.pc.in contains
> 
> Okay, my question was incorrect. I'd better ask about pkg-config. If libs.private has been supported in pkg-config for ages, then it is fine.

https://gitlab.freedesktop.org/pkg-config/pkg-config/-/blob/master/ChangeLog?ref_type=heads#L440
I guess the support of `Libs.private` started from 19 years before.

ivmai:
The patch looks good, thank you! I will process it next week.

ivmai:
Related: #247

fundawang:
Fair enough. I was also fixing guile build.

https://eur.openeuler.openatom.cn/coprs/fundawang/guile-2.2/builds/

First, I try to build gc with my patch (8.2.6-3), then build guile upon it without problem.

As I said, the core problem is how the `libatomic_ops` affects gc. If the exported symbols of libgc really relies on `libatomic_ops`, then we should use `Requires` instead of `LIBS`.

fundawang:
`Requires.private` sample patch:
https://[gitee.com/fundawang/gc/raw/master/gc-8.2.6-make-atomic_ops_private.patch](https://gitee.com/fundawang/gc/raw/master/gc-8.2.6-make-atomic_ops_private.patch)

The patch is based on 8.2.6, it does not contain changes regarding CMakeLists.txt

ivmai:
> `Requires.private` sample patch: https://[gitee.com/fundawang/gc/raw/master/gc-8.2.6-make-atomic_ops_private.patch](https://gitee.com/fundawang/gc/raw/master/gc-8.2.6-make-atomic_ops_private.patch)
> 
> The patch is based on 8.2.6, it does not contain changes regarding CMakeLists.txt

If really needed, let's do it in a separate PR. (And, yes, CMakeLists.txt should be changed too.) But, actually, I suppose, that most clients do not need libatomic_ops for bdwgc at all (because the latter relies on GCC atomic instrincs).

ivmai:
Merged.

PR #656, opened 2024-08-27
Move atomic_ops libs into Requires.private

fundawang:
Acturally, bdwgc it self may requires atomic_ops, but it does not require downstream packages (like guile) requires atomic_ops. So I would suggest that move `@ATOMIC_OPS_LIBS@` into `Requires.private`, so that won't confuse downstream packages, especially when performing check.

https://people.freedesktop.org/~dbn/pkg-config-guide.html#faq
> My library `z` uses `libx` internally, but does not expose `libx` data types in its public API. What do I put in my `z.pc` file?
> Again, add the module to `Requires.private` if it supports `pkg-config`. In this case, the compiler flags will be emitted unnecessarily, but it ensures that the linker flags will be present when linking statically. If `libx` does not support `pkg-config`, add the necessary linker flags to `Libs.private`.

So I think, move `atomic_ops` is a more appropriate solution.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -179,6 +179,7 @@ set(SRC allchblk.c alloc.c blacklst.c dbg_mlc.c dyn_load.c finalize.c
 
 set(NODIST_SRC)
 set(ATOMIC_OPS_LIBS)
+set(ATOMIC_OPS_PC)
 set(ATOMIC_OPS_LIBS_CMAKE)
 set(THREADDLLIBS_LIST)
 set(NEED_LIB_RT)
@@ -197,6 +198,7 @@ if (enable_threads)
         "with_libatomic_ops and without_libatomic_ops are mutually exclusive")
     endif()
     set(ATOMIC_OPS_LIBS "-latomic_ops")
+    set(ATOMIC_OPS_PC "atomic_ops")
     find_package(Atomic_ops CONFIG)
     if (Atomic_ops_FOUND)
       get_target_property(AO_INCLUDE_DIRS Atomic_ops::atomic_ops
--- a/configure.ac
+++ b/configure.ac
@@ -1207,7 +1207,9 @@ AS_IF([test x"$with_libatomic_ops" != xno],
   [AS_IF([test x"$with_libatomic_ops" != xnone -a x"$THREADS" != xnone],
          [AC_MSG_RESULT([external])
           ATOMIC_OPS_LIBS="-latomic_ops"
-          AC_SUBST([ATOMIC_OPS_LIBS])],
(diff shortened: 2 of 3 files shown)


ivmai:
Please rebase to fresh master and explain the difference from the previous PR I just merged.

CMakeLists.txt
```diff
@@ -197,6 +198,7 @@ if (enable_threads)
         "with_libatomic_ops and without_libatomic_ops are mutually exclusive")
     endif()
     set(ATOMIC_OPS_LIBS "-latomic_ops")
+    set(ATOMIC_OPS_PC "atomic_ops")
```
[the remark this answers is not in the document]
ivmai: Also, update comment "ATOMIC_OPS_LIBS, PACKAGE_VERSION are defined above" below

configure.ac
```diff
@@ -1207,7 +1207,9 @@ AS_IF([test x"$with_libatomic_ops" != xno],
   [AS_IF([test x"$with_libatomic_ops" != xnone -a x"$THREADS" != xnone],
          [AC_MSG_RESULT([external])
           ATOMIC_OPS_LIBS="-latomic_ops"
```
ivmai, line 1209: Define ATOMIC_OPS_LIBS based on ATOMIC_OPS_PC value.

configure.ac
```diff
@@ -1207,7 +1207,9 @@ AS_IF([test x"$with_libatomic_ops" != xno],
...
-          AC_SUBST([ATOMIC_OPS_LIBS])],
+          ATOMIC_OPS_PC="atomic_ops"
+          AC_SUBST([ATOMIC_OPS_LIBS])
+          AC_SUBST([ATOMIC_OPS_PC])],
```
ivmai, line 1212: Also define ATOMIC_OPS_PC="" below (for internal "internal" case)

ivmai (changes requested):

fundawang:
I suddenly realized a problem. `atomic_ops` shipped `atomic_ops.pc` starting from 7.2. If `bdwgc` is built upon atomic_ops 7.1, then we should not put it in `bdwgc.pc` file, because there is no `atomic_ops.pc` at that time.

Maybe we should leave it into `Libs.private` for now, till the time of `bdwgc` relies on `atomic_ops` >= 7.2.

ivmai:
> Maybe we should leave it into Libs.private for now, till the time of bdwgc relies on atomic_ops >= 7.2.

Although, 7.1 or 7.2 is not hard-coded in bdwgc source. I agree that we could leave this change for the future - I will create the relevant issue and close this PR.

PR #675, opened 2024-11-01
Export CMake libraries with aliases for easy use

jeaye:
This allows bdwgc to be included in a CMake build using `add_subdirectory`, `FetchContent`, or similar. It then relies on this CMake policy to ensure the library is found in any `target_link_libraries` usage: https://cmake.org/cmake/help/latest/policy/CMP0028.html

Without the alias, `target_link_libraries` will not be able to find `gc`, `gccpp`, etc.

More info here: https://stackoverflow.com/a/77211132

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -664,6 +664,7 @@ if (EMSCRIPTEN AND enable_emscripten_asyncify)
 endif()
 
 add_library(gc ${SRC})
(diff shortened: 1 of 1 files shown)


ivmai:
LGTM. Will be merged in a day.

ivmai:
Merged.

jeaye:
Thanks, Ivan!

PR #677, opened 2024-11-10
Minor build.zig improvements

linusg:


ivmai:
Okay, thank you, I will review and test it in several days.

ivmai:
Is this working with zig v0.12.0? If not (but working with 0.13.0) then we need to update zig_min_required_version.

linusg:
This uses no recent features and will work with Zig 0.12, yes.

ivmai:
Merged.

Issue #97, opened 2016-05-08
Support for Haiku Platform


CodeforEvolution:
Support for the operating system Haiku, the open source successor to beos.

Main Homepage:
https://www.haiku-os.org

Development Websites:
https://dev.haiku-os.org/
https://www.haiku-os.org/development

ivmai:
BeOS/x86 should be supported. Probably not tested for a long time.
Anyway, the patch is welcomed.

ivmai:
Initial support of Haiku/amd64: 847927f

Would be good to test it.

Related topic: haikuports/haikuports#2215

ivmai:
Fork handling (--enable-handle-fork) is not supported because of some bug in Haiku (as of hrev51798): pthread_mutex_unlock (called from GC_release_mark_lock called from fork_parent_proc) returns ESIGPARM.

ivmai:
Besides the above bug, there are 2 more bugs in Haiku (for which I've created workarounds in bdwgc):
* sem_wait sometimes returns EACCES (the workaround is to retry)
* pthread_join returns ESRCH if too many threads were created (sample: tests/subthread_create.c compiled with -D GC_NO_THREAD_REDIRECTS).

PR #678, opened 2024-11-25
Rework Haiku port.

waddlesplash:
Everything works except for forking from a thread other than the main thread; this causes crashes becaues the forked child's pthread_self() will be equivalent to the original process's main pthread_self() and thus the thread list is updated wrongly.

Changes:

 * Use B_CURRENT_TEAM (which is 0) in GC_register_dynamic_libraries, for clarity's sake.

 * Always use anonymous mmaps. Requesting memory from the standard libc allocator (via posix_memalign) just confuses things.

 * MPROTECT_VDB will not work correctly on anything other than recent nightlies, and also it's considerably slower than regular collecting, so disable it for now.

 * Remove the custom stack-base detection and just use pthread's. Also use this for the main thread rather than heuristics.

 * No need to search for "data start", all information should be reported correctly.

 * In GC_default_on_abort, invoke debugger(). This will cause the crash reason to appear in any debug reports generated (by the default system application crash dialog.)

 * Reinitialize the mark lock in fork_child_proc. It isn't legal to reuse any locks in child processes on Haiku; and the POSIX specification says that unlocking a mutex that a thread doesn't own is "undefined behavior" anyway.

 * Remove workarounds that don't seem to be needed anymore.

### DIFF
--- a/include/private/gcconfig.h
+++ b/include/private/gcconfig.h
@@ -895,7 +895,19 @@ extern char etext[];
 #ifdef HAIKU
 #  define OS_TYPE "HAIKU"
 #  define DYNAMIC_LOADING
-#  define MPROTECT_VDB
+#  define USE_GET_STACKBASE_FOR_MAIN
+#  undef MPROTECT_VDB		/* slow */
+#  ifndef HAVE_PTHREAD_GETATTR_NP
+#    define HAVE_PTHREAD_GETATTR_NP 1
+#  endif
+#  ifndef HAVE_CLOCK_GETTIME
+#    define HAVE_CLOCK_GETTIME 1
+#  endif
+#  ifndef USE_MMAP
+#    define USE_MMAP 1
+#  endif
+#  define USE_MMAP_ANON
+#  define USE_SPIN_LOCK
 EXTERN_C_END
 #  include <OS.h>
 EXTERN_C_BEGIN
@@ -1351,8 +1363,7 @@ extern int etext[];
 #    define STACKBOTTOM MAKE_CPTR(0x3ffff000)
 #  endif
 #  ifdef HAIKU
-extern int etext[];
-#    define DATASTART PTR_ALIGN_UP((ptr_t)etext, 0x1000)
+/* Nothing specific. */
 #  endif
 #  ifdef HURD
 /* Nothing specific. */
@@ -2247,8 +2258,7 @@ extern int _end[];
 /* Nothing specific. */
 #  endif
 #  ifdef HAIKU
-#    define HEURISTIC2
-#    define SEARCH_FOR_DATA_START
+/* Nothing specific. */
 #  endif
 #  ifdef HURD
 /* Nothing specific. */
@@ -3431,9 +3441,6 @@ void *psp2_get_mem(size_t bytes);
 #  elif defined(NINTENDO_SWITCH)
 void *switch_get_mem(size_t bytes);
 #    define GET_MEM(bytes) (struct hblk *)switch_get_mem(bytes)
-#  elif defined(HAIKU)
-ptr_t GC_haiku_get_mem(size_t bytes);
-#    define GET_MEM(bytes) (struct hblk *)GC_haiku_get_mem(bytes)
 #  elif defined(EMSCRIPTEN_TINY)
 void *emmalloc_memalign(size_t alignment, size_t size);
 #    define GET_MEM(bytes) \
--- a/os_dep.c
+++ b/os_dep.c
@@ -889,20 +889,6 @@ GC_get_stack_base(struct GC_stack_base *sb)
 #  define HAVE_GET_STACK_BASE
 #endif /* EMBOX */
 
-#ifdef HAIKU
-#  include <kernel/OS.h>
-
-GC_API int GC_CALL
-GC_get_stack_base(struct GC_stack_base *sb)
-{
-  thread_info th;
-  get_thread_info(find_thread(NULL), &th);
-  sb->mem_base = th.stack_end;
-  return GC_SUCCESS;
-}
-#  define HAVE_GET_STACK_BASE
-#endif /* HAIKU */
-
 #ifdef OS2
 GC_API int GC_CALL
 GC_get_stack_base(struct GC_stack_base *sb)
@@ -1327,7 +1313,7 @@ GC_get_main_stack_base(void)
 }
 #  define GET_MAIN_STACKBASE_SPECIAL
 
-#elif !defined(ANY_MSWIN) && !defined(EMBOX) && !defined(HAIKU) \
+#elif !defined(ANY_MSWIN) && !defined(EMBOX) \
     && !defined(OS2) && !(defined(OPENBSD) && defined(THREADS)) \
     && (!(defined(SOLARIS) && defined(THREADS)) || defined(_STRICT_STDC))
 
@@ -1397,6 +1383,8 @@ GC_get_main_stack_base(void)
   result = GC_qnx_main_stack_base();
 #    elif defined(FREEBSD_STACKBOTTOM)
   result = GC_freebsd_main_stack_base();
+#    elif defined(HAIKU)
+  ABORT("pthread_attr_getstack failed");
 #    elif defined(HEURISTIC2)
   {
     ptr_t sp = GC_approx_sp();
@@ -1428,7 +1416,7 @@ GC_get_main_stack_base(void)
   return result;
 }
 #  define GET_MAIN_STACKBASE_SPECIAL
-#endif /* !ANY_MSWIN && !HAIKU && !OS2 */
+#endif /* !ANY_MSWIN && && !OS2 */
 
 #if (defined(HAVE_PTHREAD_ATTR_GET_NP) || defined(HAVE_PTHREAD_GETATTR_NP)) \
     && defined(THREADS) && !defined(HAVE_GET_STACK_BASE)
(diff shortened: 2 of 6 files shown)


ivmai:
> Requesting memory from the standard libc allocator (via posix_memalign) just confuses things.

What does it mean? Is it just incompatible with USE_MUNMAP?

ivmai:
Also, please let me know which changes are supposed to be of "fix" or improvement" kind. My guess is:

> Always use anonymous mmaps. Requesting memory from the standard libc allocator (via posix_memalign) just confuses things.

Unclear to me, see my question above.

> MPROTECT_VDB will not work correctly on anything other than recent nightlies, and also it's considerably slower than regular collecting, so disable it for now.

A fix.

> Remove the custom stack-base detection and just use pthread's. Also use this for the main thread rather than heuristics.

Unclear to me.

> No need to search for "data start", all information should be reported correctly.

Unclear to me.

> In GC_default_on_abort, invoke debugger(). This will cause the crash reason to appear in any debug reports generated (by the default system application crash dialog.)

Improvement.

> Reinitialize the mark lock in fork_child_proc. It isn't legal to reuse any locks in child processes on Haiku; and the POSIX specification says that unlocking a mutex that a thread doesn't own is "undefined behavior" anyway.

A fix.

> Remove workarounds that don't seem to be needed anymore.

Improvement (unless you see a clear relation between a workaround to any of the above fixes).

waddlesplash:
> What does it mean? Is it just incompatible with USE_MUNMAP?

At least this, but I was seeing strange crashes with the old code regardless. So I would call it a "Fix".

> Remove the custom stack-base detection and just use pthread's. ...

For threads other than the main, this is just an "improvement", since the two methods yield the same results. For the main thread this is a fix, since before we were not getting the stack base correctly, I believe.

> No need to search for "data start", all information ...

Probably just an improvement, since all data regions will be reported in scanning for images, so we don't need to search for them.

The rest of your categorizations seem correct.

ivmai:
//cc @NattyNarwhal (the author of initial port from mono libgc - commit 847927f0e)

include/private/gcconfig.h
```diff
@@ -895,7 +895,19 @@ extern char etext[];
...
+#    define HAVE_PTHREAD_GETATTR_NP 1
+#  endif
+#  ifndef HAVE_CLOCK_GETTIME
+#    define HAVE_CLOCK_GETTIME 1
+#  endif
+#  ifndef USE_MMAP
+#    define USE_MMAP 1
+#  endif
+#  define USE_MMAP_ANON
+#  define USE_SPIN_LOCK
```
ivmai, line 910: > Interestingly with USE_SPIN_LOCK it fails even faster; instead of taking a handful of seconds (with low CPU usage) before crashing, it crashes very quickly after starting with the "Unexpected segmentation fault outside heap ...".

Let me not define USE_SPIN_LOCK. A comment near define USE_SPIN_LOCK in gc_lock.h says:
```
/* In the THREAD_LOCAL_ALLOC case, the allocator lock tends to    */
/* be held for long periods, if it is held at all.  Thus spinning */
/* and sleeping for fixed periods are likely to result in         */
/* significant wasted time.  We thus rely mostly on queued locks. */
```
I think a more detailed investigation is needed whether to use spin locks (and not only for Haiku).

os_dep.c
```diff
@@ -2251,7 +2239,7 @@ GC_register_data_segments(void)
   /* Avoid even referencing DATASTART and DATAEND as they are   */
   /* unnecessary and cause linker errors when bitcode is        */
   /* enabled.  GC_register_data_segments is not called anyway.  */
-#  elif defined(DYNAMIC_LOADING) && defined(DARWIN)
+#  elif defined(DYNAMIC_LOADING) && (defined(DARWIN) || defined(HAIKU))
   /* No-op.  GC_register_main_static_data() always returns false. */
```
ivmai, line 2243: GC_register_main_static_data() return constant false only for Darwin and some FreeBSD variants (unless USE_PROC_FOR_LIBRARIES). But not for Haiku.

include/private/gcconfig.h
waddlesplash: It doesn't fail at all without fork(), it works fine with the spin lock, and it's much faster than without (I think 1-2 seconds vs. over 10 for gctest, with vs. without spinlocks.) The spinlock code falls back to a queued lock after a number of spins fail, it appears?

On Haiku, pthread locks never spin but always call into the kernel if the lock can't be acquired with a simple atomic operation. On other OSes, some spins may be used ("PTHREAD_MUTEX_ADAPTIVE" or the like.) `USE_SPIN_LOCK` seems to replicate that behavior.

os_dep.c
waddlesplash: On Haiku, the main application binary is also a "shared image" and will be reported in an `image_info` same as any dynamically loaded libraries.

include/private/gcconfig.h
ivmai: > it's much faster than without (I think 1-2 seconds vs. over 10 for gctest, with vs. without spinlocks.)

I see. Would be nice if you could provide a table of gctest run time (e.g. the one printed by "testing done in %lu ms") of 10-20 runs with and w/o USE_SPIN_LOCK. (I your PC is too powerful, you could increase NTHREADS to make each run longer to reduce noise.)

os_dep.c
ivmai: OK, I see. Then we should define `GC_register_main_static_data() { return FALSE; }` (like in Darwin case).

pthread_support.c
```diff
@@ -1510,12 +1510,7 @@ fork_child_proc(void)
 #    endif
 #    ifdef PARALLEL_MARK
   if (GC_parallel) {
-#      if defined(THREAD_SANITIZER) && defined(GC_ASSERTIONS) \
```
ivmai, line 1513: Will it be OK if the existing pthread_mutex_unlock/GC_release_mark_lock call is left as-is, but after it we call pthread_mutex_init?

ivmai:
> Reinitialize the mark lock in fork_child_proc. It isn't legal to reuse any locks in child processes on Haiku; and the POSIX specification says that unlocking a mutex that a thread doesn't own is "undefined behavior" anyway.

Related [comment](https://github.com/ivmai/bdwgc/issues/97#issuecomment-370918797) in issue #97.

ivmai:
> Remove workarounds that don't seem to be needed anymore.

Related comment: https://github.com/ivmai/bdwgc/issues/97#issuecomment-370919986

* https://github.com/ivmai/bdwgc/commit/47ca7611ee3a0946346728ccf0125f3682af9d63 - let me keep it for now
* https://github.com/ivmai/bdwgc/commit/1d9c1e437ba41bcfcb94489b4d12bbf8aaa9d056 - okay, I will revert it

pthread_support.c
waddlesplash: It will still error, but since the error isn't checked at the moment except in assert builds, it isn't causing any problems. Ideally it should not be unlocked at all but reinitialized immediately.

ivmai:
I have merged most of your changes (except for DATASTART-related and USE_SPIN_LOCK ones, for now), with some refactoring.
Please test the current master.

pthread_support.c
ivmai: > Ideally it should not be unlocked at all but reinitialized immediately.

The manual (https://man7.org/linux/man-pages/man3/pthread_mutex_destroy.3p.html) says the opposite:
- Attempting to destroy a locked mutex, or a mutex that another thread is attempting to lock, ..., results in undefined behavior
- Attempting to initialize an already initialized mutex results in undefined behavior

> It will still error, but since the error isn't checked at the moment except in assert builds, it isn't causing any problems.

Right, the result is cast to void intentionally (both for `mark_mutex` and `GC_allocate_ml`).
I will remove checking in the assertion, thank you!

waddlesplash:
Well, out of the box it doesn't build:

```
extra/../os_dep.c: In function 'GC_get_main_stack_base':
extra/../os_dep.c:1404:8: error: #error None of HEURISTIC* and *STACKBOTTOM defined!
 1404 | #      error None of HEURISTIC* and *STACKBOTTOM defined!
      |        ^~~~~
Makefile:1719: recipe for target 'extra/gc.lo' failed
```

But it seems this is due to a small change I missed adding to this patch:
```diff
diff --git a/os_dep.c b/os_dep.c
index 339a1566..ebdbfd58 100644
--- a/os_dep.c
+++ b/os_dep.c
@@ -1383,6 +1383,8 @@ GC_get_main_stack_base(void)
   result = GC_qnx_main_stack_base();
 #    elif defined(FREEBSD_STACKBOTTOM)
   result = GC_freebsd_main_stack_base();
+#    elif defined(HAIKU)
+  ABORT("pthread_attr_getstack failed");
 #    elif defined(HEURISTIC2)
   {
     ptr_t sp = GC_approx_sp();
```
After applying the above (and building with `NO_TEST_HANDLE_FORK`), it compiles and runs, completing with:
```
Collector appears to work (testing done in 14671 ms)
```
And if I enable `USE_SPIN_LOCK`:
```
Collector appears to work (testing done in 2286 ms)
```

ivmai:
Okay, I will update

ivmai:
> After applying the above

Does this also work if `--disable-threads` is passed to configure?

waddlesplash:
No, but that message isn't the one hit, instead we get:
```
Running on HAIKU/X86_64 target
Supported VDBs: manual default
GC_is_visible produced wrong failure indication
Test failed
```

ivmai:
> GC_is_visible produced wrong failure indication

This means there is some other bug, but in multithread builds it is not observed because of the relaxed checking in GC_is_visible.

ivmai:
> os_dep.c:1404:8: error: #error None of HEURISTIC* and *STACKBOTTOM defined!

Should be fixed now, please test it again.

ivmai:
> > GC_is_visible produced wrong failure indication
> 
> This means there is some other bug, but in multithread builds it is not observed because of the relaxed checking in GC_is_visible.

Let's move this to a separate issue: #680

waddlesplash:
> Should be fixed now, please test it again.

Works fine, thanks.

Also, I retested single-threaded GC, and it also works now. Perhaps because of the changes to find DATASTART and the like?

ivmai:
Closing this PR as done. Other fixes to be done in other issues.
I will cherry-pick this PR to release-8_2 branch later (in a month).

ivmai:
@waddlesplash, please test https://github.com/ivmai/bdwgc/tree/release-8_2 branch, this PR commits are now backported to this branch. The upcoming stable release (v8.2.10) will be based on this branch.

waddlesplash:
All tests still pass including gctest after fork.

Issue #627, opened 2024-03-20
Support CHERI extension (port patches from capablevms org)


ivmai:
The purpose of this is not only support the protected mode but also better test bdwgc.

The patches in https://github.com/capablevms/bdwgc (master) should be useful.

Related issue: #461

ivmai:
Current status: I've done a lot of refactoring to simplify porting to CHERI, and backported some changes from https://github.com/capablevms/bdwgc/ (see c29365b). Still many CHERI-specific changes to be backported. And below is the list of code places worth attention during enabling of CHERI in master:
* include/gc/gc.h: GC_reachable_here: should use "g" constraint
* include/gc/gc.h: GC_HIDE_POINTER, GC_HIDE_NZ_POINTER (and the reveal ops)
* include/gc/gc_tiny_fl.h: GC_TINY_FREELISTS (is the current value OK?)
* include/private/dbg_mlc.h: GC_FREED_MEM_MARKER, START_FLAG, END_FLAG
* include/private/gcconfig.h: PTR_ALIGN_DOWN, PTR_ALIGN_UP: use cheri_align_down/up
* include/private/gcconfig.h: RISCV: define CPP_PTRSZ to __riscv_clen
* include/private/gcconfig.h: FIXUP_POINTER
* tests/gctest.c: check_heap_stats: increase max_heap_sz if CPP_PTRSZ > 64

ivmai:
And below is the list of code places worth attention during enabling of CHERI in master:
* include/gc/gc.h: GC_ADDR_LT: compare addr part only?
* include/gc/gc.h: GC_reachable_here: should use "g" constraint
* include/gc/gc.h: GC_HIDE_POINTER, GC_HIDE_NZ_POINTER (and the reveal ops)
* include/gc/gc_tiny_fl.h: GC_TINY_FREELISTS (is the current value OK?)
* include/private/dbg_mlc.h: GC_FREED_MEM_MARKER, START_FLAG, END_FLAG
* include/private/gcconfig.h: PTR_ALIGN_DOWN, PTR_ALIGN_UP: use cheri_align_down/up
* include/private/gcconfig.h: RISCV: define CPP_PTRSZ to __riscv_clen
* include/private/gcconfig.h: FIXUP_POINTER
* tests/gctest.c: check_heap_stats: increase max_heap_sz if CPP_PTRSZ > 64

ivmai:
[...]

PR #695, opened 2025-01-15
Disable contiguous mmap()  page suggestions for CHERI platforms

djichthys:
1. Disable contiguous memory page maps from the OS in CHERIBSD. Coalescing is disabled for CHERI purecap due to capability *out-of-bounds* errors. Done to eliminate errors from mmap() because:  
    * CHERI purecap implementations of mmap() on CHERIBSD return error on requesting memory with a suggested capability. 
    * A fully valid capability is required and this cannot be provided ahead of time within the bdwgc library. 
2. Also when reclaiming memory while coalescing, utilise a capability rather than a computed pointer value -- otherwise capability provenance is lost.


os_dep.c
```diff
@@ -2387,7 +2387,14 @@ GC_unix_mmap_get_mem(size_t bytes)
     return GC_unix_mmap_get_mem(bytes);
   }
 #      endif
+
+#      ifndef CHERI_PURECAP
   last_addr = PTR_ALIGN_UP((ptr_t)result + bytes, GC_page_size);
+#      else
+  /* CHERI_PURECAP: Non-NULL addresses don't work in CheriABI */
+  last_addr = NULL;
```
ivmai, line 2395: I've already fixed this in a different way - use address for the hint instead of the pointer (with capabilities): https://github.com/ivmai/bdwgc/commit/5fe8a391eb3d7c13a44297806734dff6989dc58d

headers.c
```diff
@@ -430,7 +430,11 @@ GC_prev_block(struct hblk *h)
         j -= (GC_signed_word)ADDR(hhdr);
       } else {
         /* TODO: return hhdr -> hb_block instead */
+#ifndef CHERI_PURECAP
```
ivmai, line 433: Unfortunately this helps only with small tests. E.g. gctest fails at some moment because hhdr->hb_block is null (but it should not be null). A mentioned in the TODO above, we should return hb_block for all platforms but to do this we need to fix setting hb_block in allchblk.c. (I'm struggling with it these days but have not found a solution yet.)

ivmai:
@djichthys, thank your great work starting from 2021! I hope I've integrated all your changes a long with serious code base refactoring, see #627 (I have not integrated Richards test to avoid a license conflict).

ivmai:
Let me close this PR because in overall (after merging this the fresh master) contains only single line change (about `hb_block`) but the change does not solve the entitled failure.

ivmai:
> contains only single line change (about `hb_block`) but the change does not solve the entitled failure.

This should be now fixed by 816a802b0

PR #705, opened 2025-02-15
Avoid GC_suspend aborting due to deleted thread

### LINKED ISSUE #704: GC_delete_thread call from GC_DllMain interleaves with GC_suspend from GC_gcollect causing SuspendThread error
Hi, I'm debugging an issue with the [neko vm](https://github.com/HaxeFoundation/neko) which makes use of bdwgc. (For full context, I'm using a [slightly modified neko](https://github.com/tobil4sk/neko/tree/dc69e3fda1810a36d62e766eca9bdd5b782ba94a) updated to 8.2.8).

On Windows, it looks like occasionally when a thread has been created and eventually finishes running, the [DLLMain call to detach and delete that thread](https://github.com/ivmai/bdwgc/blob/db6668de9406f9d9642d1ad520a153c03b65c512/win32_threads.c#L3253) can be interleaved with a `GC_gcollect()` call in neko. The `GC_gcollect` still sees the thread so it tries to suspend it, which results in an invalid handle error.

Neko enables thread discovery in bdwgc on windows. These are the cmake flags bdwgc was built with:
```
-Wno-dev
-Denable_threads=ON
-Denable_thread_local_alloc=OFF
-DCMAKE_USE_WIN32_THREADS_INIT=ON
-DCMAKE_CXX_STANDARD=14
```

It seems like once one thread starts using a table entry, no other thread should access that entry? I.e. once a thread is in the process of running `GC_delete_thread`, it should not be possible for another thread to begin trying to run `GC_suspend`. Does this sound like it may be a genuine problem with bdwgc? Or is it more likely to be an issue with how neko is using it?

tobil4sk:
A thread may be deleted by a DllMain call during execution of GC_suspend.

This patch sets the handle to be NULL when deleting the thread to avoid use-after-free bugs, and skips the GC_suspend attempt if the thread handle has been closed.

Fixes #704

### DIFF
--- a/pthread_support.c
+++ b/pthread_support.c
@@ -827,7 +827,9 @@ GC_INNER_WIN32THREAD void
 GC_delete_thread(GC_thread t)
 {
 #  if defined(GC_WIN32_THREADS) && !defined(MSWINCE)
-  CloseHandle(t->handle);
+  HANDLE handle = t->handle;
+  t->handle = NULL;
+  CloseHandle(handle);
 #  endif
 #  if !defined(GC_NO_THREADS_DISCOVERY) && defined(GC_WIN32_THREADS)
   if (GC_win32_dll_threads) {
--- a/win32_threads.c
+++ b/win32_threads.c
@@ -431,7 +431,7 @@ GC_suspend(GC_thread t)
     GC_acquire_dirty_lock();
   }
 #  elif defined(RETRY_GET_THREAD_CONTEXT)
-  for (retry_cnt = 0;;) {
+  for (retry_cnt = 0; t->handle != NULL;) {
     /* Apparently the Windows 95 GetOpenFileName call creates         */
     /* a thread that does not properly get cleaned up, and            */
     /* SuspendThread on its descriptor may provoke a crash.           */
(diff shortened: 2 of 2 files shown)


ivmai:
Looking into the patch...

ivmai:
I've added barriers to operate with t->handle and done a bit of refactoring of your patch. Please test it.
(Later I will back-port it to release-8_2 branch.)

tobil4sk:
Thanks for having a look! Your version of the patch seems to work well. I still have some sort of deadlock that occurs during dllmain calls very occasionally, but this occurred with my patch too. I guess it may be related to: #707.

ivmai:
Note that issue #707 is not related to dllmain (because shared build is off).

tobil4sk:
> Note that issue https://github.com/ivmai/bdwgc/issues/707 is not related to dllmain (because shared build is off).

I guess it may be a different issue then, I've opened: #710

tobil4sk:
> (Later I will back-port it to release-8_2 branch.)

I backported the patch because I wanted to test the release-8_2 branch with neko, so I've put it in #711 in case it helps

ivmai:
Okay, I will review it in a week. Thank you!

PR #706, opened 2025-02-19
Stop using removed std.Target functions in build.zig

linusg:
See: https://github.com/ziglang/zig/commit/481b7bf3f095488a89e20d88ada092529bc6e6f8

This change is backwards compatible because the old functions are equivalent to the new code.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -45,7 +45,7 @@ pub fn build(b: *std.Build) void {
     const target = b.standardTargetOptions(.{});
     const t = target.result;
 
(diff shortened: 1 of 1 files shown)


ivmai:
> This change is backwards compatible because the old functions are equivalent to the new code.

It works with zig 0.12.0, right?

linusg:
It should, yes. That's ancient though, I wouldn't recommend using anything older than 0.13 - 0.14 is on the horizon.

ivmai:
Merged. Thank you!

PR #711, opened 2025-03-07
Backport fix for #705 to release-8_2

### LINKED ISSUE #704: GC_delete_thread call from GC_DllMain interleaves with GC_suspend from GC_gcollect causing SuspendThread error

tobil4sk:
Backports these two commits (which fix #704) to the release-8_2 branch:
- 553f9f0aebe677fab82c3a0bfbff03325fc72854
- 5658d91c8dd3d9d7cb19bc447d5b0a9f4d4563f2

### DIFF
--- a/win32_threads.c
+++ b/win32_threads.c
@@ -704,14 +704,15 @@ GC_API void GC_CALL GC_register_altstack(void *stack GC_ATTR_UNUSED,
 /* GC_win32_dll_threads is set.  Does not actually free         */
 /* GC_thread entry (only unlinks it).                           */
 /* If GC_win32_dll_threads is set it should be called from the  */
-/* thread being deleted.                                        */
+/* thread being deleted (except for DLL_PROCESS_DETACH case).   */
 STATIC void GC_delete_gc_thread_no_free(GC_vthread t)
 {
-# ifndef MSWINCE
-    CloseHandle(t->handle);
-# endif
 # ifndef GC_NO_THREADS_DISCOVERY
     if (GC_win32_dll_threads) {
+      HANDLE handle = t->handle;
+
+      AO_store_release((volatile AO_t *)&t->handle, NULL);
+      CloseHandle(handle);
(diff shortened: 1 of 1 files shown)


ivmai:
Thank you!
Commit 1: I am not going to back port containing only documentation (and the comment is not relevant to the issue)
Commit 2: Looks good (the only my remark is that NULL -> 0). I've adjusted the patch and will push to github soon (after local testing).

tobil4sk:
Makes sense, thank you!

ivmai:
Merged to release-8_2 branch (along with some other fixes).

Issue #710, opened 2025-03-07
GC_suspend can hang attempting to suspend DllMain thread


tobil4sk:
It seems that in some cases, the SuspendThread call within GC_suspend may hang and never return if it is trying to suspend the thread that is running DllMain to detach itself: https://github.com/ivmai/bdwgc/blob/b435e76287dbf275d922a0a3c432efc913c8460e/win32_threads.c#L457

This seems related to the previously fixed issue: #704. Again I am running bdwgc in the neko vm on Windows with DllMain enabled, with this commit: 5658d91c8dd3d9d7cb19bc447d5b0a9f4d4563f2, modified with a few debug print statements.

tobil4sk:
I was previously debugging by adding printf statements, however, I have now run a proper debugger on the program and removed the print statements to avoid their potential side effects. It seems that without modification, gc.dll actually hangs on `GC_win32_MessageBoxA`, which it attempts to run due to `ResumeThread failed in suspend loop`. This error happens because it runs `ResumeThread` on a NULL handle. It looks like between the SuspendThread and the ResumeThread, the thread handle has started closing the thread and set it to NULL.

I'm not sure why `MessageBoxA` hangs since the message box isn't actually showing, but I guess GC_suspend shouldn't abort on the ResumeThread failure if it is caused by a NULL handle.

ivmai:
Should be fixed now (but introducing a potential resource leak as mentioned in https://github.com/ivmai/bdwgc/pull/712#issuecomment-2725616155).

ivmai:
Let me keep it open as the current solution has some drawback and need to be reworked.

ivmai:
On behalf of @tobil4sk (copied from PR #712):
It seems that in these cases where we suspend a thread running DllMain and lose the handle so are unable to resume, it can prevent new threads from being able to enter DllMain at all, so attempting to create a new thread doesn't work and the thread never starts running.

ivmai:
On behalf of @tobil4sk (copied from PR #712):
Alternatively, maybe DllMain shouldn't run CloseHandle at all, but instead mark the thread with some flag so that it gets closed by the gc thread the next time it iterates through the thread table?

tobil4sk:
I managed to strip down my neko test program down to a basic C sample.

[...]

PR #712, opened 2025-03-07
Avoid GC_suspend error if DllMain closes handle

tobil4sk:
It's possible that DllMain has set the thread handle to NULL since the SuspendThread call and that is why ResumeThread has failed. In this case, GC_suspend shouldn't abort.

This is especially problematic as it seems that sometimes the program hangs without displaying the message box. See: https://github.com/ivmai/bdwgc/issues/710#issuecomment-2707608252

### DIFF
--- a/win32_threads.c
+++ b/win32_threads.c
@@ -472,8 +472,15 @@ GC_suspend(GC_thread t)
       }
 
       /* Resume the thread, try to suspend it in a better location.   */
-      if (ResumeThread(t->handle) == (DWORD)-1)
+      if (ResumeThread(t->handle) == (DWORD)-1) {
+#    ifndef GC_NO_THREADS_DISCOVERY
+        if (NULL == GC_cptr_load_acquire(&t->handle)) {
+          GC_release_dirty_lock();
+          return;
+        }
+#    endif
         ABORT("ResumeThread failed in suspend loop");
+      }
     } else {
 #    ifndef GC_NO_THREADS_DISCOVERY
       if (NULL == GC_cptr_load_acquire(&t->handle)) {


ivmai:
I don't understand how could we get error on ResumeThread, or to say more precisely how could DLLMain could be executed for a thread which is suspended.

tobil4sk:
This is just speculation, as it is quite difficult to debug this without introducing side effects, but maybe something like this could be happening?
1) GC_suspend calls SuspendThread on a valid handle
2) Within the SuspendThread call a context switch occurs to DllMain (before the thread has actually been suspended)
3) DllMain sets t->handle to NULL, but does not yet close the handle
4) A context switch occurs returning to SuspendThread which completes on the handle that was originally passed into it
5) Then ResumeThread attempts to run on t->handle which is now NULL, which gives an error causing GC_suspend to abort

ivmai:
I wonder what will happen to thread if we fail on resume thread when handle is zero? Will the thread remain suspended forever?

win32_threads.c
```diff
@@ -472,8 +472,15 @@ GC_suspend(GC_thread t)
       }
 
       /* Resume the thread, try to suspend it in a better location.   */
-      if (ResumeThread(t->handle) == (DWORD)-1)
+      if (ResumeThread(t->handle) == (DWORD)-1) {
```
ivmai, line 475: If we handle ResumeThread here, I guess we might potentially get same error in GC_start_world, and we might get same error returned by GetThreadContext in GC_push_stack_for.

ivmai (changes requested):

tobil4sk:
> I wonder what will happen to thread if we fail on resume thread when handle is zero? Will the thread remain suspended forever?

I suppose that could happen. We'd be left with no way to resume it since the handle has been set to NULL, and it will never clean itself up because it has been suspended before reaching CloseHandle...

Maybe this can avoided by using DuplicateHandle somehow. This might avoid the issue of trying to do something using a handle that another thread has closed.

Alternatively, maybe DllMain shouldn't run CloseHandle at all, but instead mark the thread with some flag so that it gets closed by the gc thread the next time it iterates through the thread table?

ivmai:
>> Will the thread remain suspended forever?
>
> Maybe this can avoided by ..

Another potential way, I think, is to always call ResumeThread before CloseHandle in DLLMain.

ivmai:
Okay, let's just put WARN() there for now - at least WARN is better than ABORT.

ivmai:
> Alternatively, maybe DllMain shouldn't run CloseHandle at all, but instead mark the thread with some flag so that it gets closed by the gc thread the next time it iterates through the thread table?

Let's do in a separate PR, afte this PR is merged.

tobil4sk:
> Another potential way, I think, is to always call ResumeThread before CloseHandle in DLLMain.

I think there are always cases where this could lead to problems, e.g.:

- GC thread calls SuspendThread
- Context switch occurs within the SuspendThread call prior to completion
- DllMain thread makes the thread handle NULL and calls ResumeThread, which fails since the thread is not suspended
- Context switch occurs back into SuspendThread, which completes leaving the Thread in a suspended state.

Now the DllMain thread will never reach CloseHandle since it is suspended.

> Okay, let's just put WARN() there for now - at least WARN is better than ABORT.

Where should the WARN run? For all previous ABORT usages after ResumeThread/GetThreadContext, including in GC_suspend?

ivmai:
Okay, let me polish current PR and merge it first.

win32_threads.c
ivmai: Fixed in c1a45b9ad

ivmai:
> Okay, let me polish current PR and merge it first.

Done. Please test it.

tobil4sk:
> Done. Please test it.

Thanks! It still works to avoid the ResumeThread issue.

It seems that in these cases where we suspend a thread running DllMain and lose the handle so are unable to resume, it can prevent new threads from being able to enter DllMain at all, so attempting to create a new thread doesn't work and the thread never starts running.

ivmai:
Got it. So, it seems no reason to cherry pick the current fix to release-8_2 branch as more work is needed.

ivmai:
Let's continue investigation in #710

PR #637, opened 2024-04-23
Zig: MacOS SDK for cross-compile

kassane:
e.g.: Build Linux to MacOS (both x86_64)

### Fix

```bash
zig build --summary all -Dtarget=native-macos
install
└─ install gc
   └─ zig build-lib gc Debug native-macos 1 errors
/home/kassane/bdwgc/extra/../os_dep.c:4453:10: error: 'mach/exception.h' file not found
#include <mach/exception.h>
         ^~~~~~~~~~~~~~~~~~~
/home/kassane/bdwgc/extra/gc.c:62:10: note: in file included from /home/kassane/bdwgc/extra/gc.c:62:
#include "../os_dep.c"
         ^
```

cc: @ivmai 

### Reference
- https://github.com/ziglang/zig/issues/18257

PR #718, opened 2025-03-21
Update build.zig file to Zig 0.14.0

congcongkuo:
According to Zig 0.14.0 [release note](https://ziglang.org/download/0.14.0/release-notes.html#addLibrary-Function), `addSharedLibrary` and `addStaticLibrary` are deprecated and should be replaced by `addLibrary(.{.linkage = linkage,})`.


ivmai:
Okay. let me review the patch. Thank you!

ivmai:
Hello @linusg. It would be nice if you could participate in the review.

build.zig
```diff
@@ -496,18 +502,26 @@ pub fn build(b: *std.Build) void {
 
     var gccpp: *std.Build.Step.Compile = undefined;
     var gctba: *std.Build.Step.Compile = undefined;
+    const gccpp_mod = b.createModule(.{
```
linusg, line 505: I'd suggest moving this into the conditional block where it's actually needed

build.zig
```diff
@@ -496,18 +502,26 @@ pub fn build(b: *std.Build) void {
...
+        .target = target,
+        .optimize = optimize,
+    });
+    const gctba_mod = b.createModule(.{
```
linusg, line 509: Likewise

build.zig
```diff
@@ -548,18 +562,22 @@ pub fn build(b: *std.Build) void {
     }
 
     var cord: *std.Build.Step.Compile = undefined;
+    const cord_mod = b.createModule(.{
```
linusg, line 565: Likewise

linusg (changes requested):

build.zig
ivmai: Yes, agree (for all).

ivmai:
Hmm. After the change zig-0.12 no longer works. OTOH w/o this change zig-0.14 works fine. So, why do we need this change?

linusg:
These methods are marked as deprecated and will be removed during the 0.15 development cycle (which is now). Backwards compatibility with more than one minor release of Zig is not desirable IMO, it's moving too quickly for that.

build.zig
```diff
@@ -449,6 +454,7 @@ pub fn build(b: *std.Build) void {
 
     // TODO: as of zig 0.12, exception.h and getsect.h are not provided
     // by zig itself for Darwin target.
+    // See https://github.com/ziglang/zig/issues/18257
```
ivmai, line 457: I don't like adding URLs to source code, this URL exists in the PR #637. Instead, let's update "as of zig 0.12" to "as of zig 0.14".

ivmai: Also, since this commit is positioned as "update to 0.14", it would be good to revise all mentions of "0.12" in this file and either update to "0.14" or remove the workaround if no longer needed.

ivmai:
Meanwhile I've updated version to 0.14 in github workflow. See commit fb254c14

ivmai:
A quick question: "Zig build system and Zig ships with an embedded clang. As a consequence, we do not have to support lots of different compilers." Is this still true (as of 0.14.0)?

ivmai:
> These methods are marked as deprecated and will be removed during the 0.15 development cycle (which is now). Backwards compatibility with more than one minor release of Zig is not desirable IMO, it's moving too quickly for that.

Zig 0.13 does not have addLibrary... Okay, let upgrade to min_version to 0.14 now. @congcongkuo, please resolve mentioned issues.
Thank you!

linusg:
> A quick question: "Zig build system and Zig ships with an embedded clang. As a consequence, we do not have to support lots of different compilers." Is this still true (as of 0.14.0)?

Yes, Zig still bundles LLVM (i.e. clang) and that's not going to change - long term the plan is to make that a separate optional component that can be pulled in for Zig projects with C dependencies (see https://github.com/ziglang/zig/issues/16270) but that's in the future and would probably still mean one known version of LLVM.

ivmai:
https://github.com/ivmai/bdwgc/actions/runs/13999602328
A lot of errors when cross compiling

linusg:
Likely a result of https://github.com/ziglang/zig/commit/11915ae48e44cfd9d4e5c5823919fb24951e68f9 and similar changes, should be easy to resolve. I'm successfully cross-compiling bdwgc for numerous targets with Zig 0.14 in my project (see https://codeberg.org/kiesel-js/kiesel/src/branch/main/.forgejo/workflows/ci.yml).

congcongkuo:
> @congcongkuo, if you have time, please try to resolve the cross-compilation issues. Please rebase your code to latest master.

maybe done

build.zig
congcongkuo: OK, I have removed it.

build.zig
congcongkuo: done

congcongkuo:
``` Zig
if (build_cord and enable_werror and !enable_threads
        and (t.abi == .gnueabi or t.abi == .gnueabihf or t.abi == .musleabi
             or t.abi == .musleabihf)) {
        // TODO: as of zig 0.12, if GCC built-in atomic intrinsic is used,
        // "large atomic operation may incur significant performance penalty"
        // warning is reported for 32-bit arm targets.
        flags.append("-D AO_DISABLE_GCC_ATOMICS") catch unreachable;
 }
```

I think I can't check whether this bug still exists, it's out of my ability.

congcongkuo:
There are also 2 comments related to Zig 0.12 that only affect MSVC build. I will try to check them lately.

congcongkuo:
> @congcongkuo, if you have time, please try to resolve the cross-compilation issues. Please rebase your code to latest master.

- [x] Failures on MIPS should have been fixed
- [ ] `thumb-linux-gnueabi` is removed [.](https://github.com/ziglang/zig/commit/281707acf3f796e624e5faf5e338878050909f27) Maybe I can simply remove it.
- [ ] `powerpc64-linux-gnu` I need help.
- [ ] `zig build -Dtarget=x86-linux-musl -DBUILD_SHARED_LIBS=true` has issue, but `zig build -Dtarget=x86-linux-musl -DBUILD_SHARED_LIBS=false` is ok. I also need some help on this.

ivmai:
> ... is removed ziglang. Maybe I can simply remove it?

Yes, just please remove in this case.

ivmai:
I've approved GH actions in this PR, just started. We actully interested in 2 zig actions.

ivmai:
> powerpc64-linux-gnu I need help.

I will look at it on Monday.

congcongkuo:
Oops! I am sorry. I wrote an error spelling of `mipsel-linux-musleabi`. And I think there are somy nontrivial problem on `mips64-linux-muslabi64` and `mips64el-linux-muslabi64`.

congcongkuo:
There is a potential improvement of `build.zig`. Currently `build.zig` uses `const build_shared_libs = b.option(bool, "BUILD_SHARED_LIBS", "Build shared libraries (otherwise static ones)") orelse true;`. This maybe replaced with a standard zig build system option `const linkage = b.option(std.builtin.LinkMode, "linkage", "Link mode for a foo_bar library") orelse .dynamic;`. I can provide 3 strategy.

1. Replaced with standard zig build system option. Pros: It can provide uniform CLI to end user and I can remove some `if` statement from `build.zig` that chenk static/shared to make it more clean.  Cons: This will break current workflow. We need to fix CI script.
2. Keeping current option but use `std.builtin.LinkMode`  internally. Pros: We don't break any interface but we can remove some `if` statement from `build.zig`. Cons: We don't provide a uniform CLI comparing to other projects using Zig build system.
3. Don't do anything.

ivmai:
Could we decide and do potential improvement in a separate PR? After we fix issues in this PR to update to 0.14

congcongkuo:
> Could we decide and do potential improvement in a separate PR? After we fix issues in this PR to update to 0.14

OK

congcongkuo:
When checking `zig-cross-build.yml`, I found some excluded build in  can bow be built successful. Should I remove these exclude rules? @ivmai

ivmai:
> When checking `zig-cross-build.yml`, I found some excluded builds can now be built successful. Should I remove these exclude rules?

Yes! Thank you for noticing this!

Also, please recheck the FIXME about build_cord+shared_libs+native-native-msvc in zig-build.yml

ivmai:
>  I think there are some nontrivial problem

Actually, it is OK to comment out (adding a FIXME) failing triplets if it is non-trivial to fix it. So, that not to block the current PR for too long.

ivmai:
> ```zig
> if (build_cord and enable_werror and !enable_threads
>         and (t.abi == .gnueabi or t.abi == .gnueabihf or t.abi == .musleabi
>              or t.abi == .musleabihf)) {
>         // TODO: as of zig 0.12, if GCC built-in atomic intrinsic is used,
>         // "large atomic operation may incur significant performance penalty"
>         // warning is reported for 32-bit arm targets.
>         flags.append("-D AO_DISABLE_GCC_ATOMICS") catch unreachable;
>  }
> ```
> 
> I think I can't check whether this bug still exists, it's out of my ability.

Please check if removal of this workaround (addition of this flag) breaks the relevant builds - if not then remove the workaround.

ivmai:
> I have checked all builds. Unfortunately all problems related to MSVC are still there. The last thing need to be resolved is memory leaking on a ubuntu native build.

Okay. Would be nice if you could add links here to the relevant bug reports in zig issues.

Thank you!

ivmai:
BTW. Are there any new targets supported by zig since 0.12? It would be nice to add them to crosscompile yml file.

congcongkuo:
Maybe loongarch? I’m no familier with it.

ivmai:
> I remove this workaround, and all builds related to it are fine. Do I need some additional check?

I think no.

ivmai:
> Maybe loongarch? I’m no familiar with it.

arc, riscv32, sparc (32-bit), loongarch64. These had been tested with gnu make. The question is if they are now supported (just compilation passes).

build.zig
ivmai: I will move it myself

congcongkuo:
I think this is better and just a slight change. What do you think of it? @ivmai 
or another pr to talk about it

ivmai:
> change

I think we could do it in this PR

.github/workflows/zig-build.yml
```diff
@@ -33,6 +33,16 @@ jobs:
         - build_cord: true # FIXME: cordtest hangs if shared lib build by MSVC
           shared_libs: true
           ttriple: native-native-msvc
+        - os: ubuntu-latest # FIXME: gctest or threadleaktest hang
```
ivmai, line 36: Let me not exclude this config (it is not related to Zig)

ivmai:
Merged.

ivmai:
> > change
> 
> I think we could do it in this PR

No, do other improvements in other PR please.

ivmai:
> Another suggestion - instead of this:
> 
> ```zig
> const gc = if (build_shared_libs) blk: {
>     // TODO: convert VER_INFO values to [SO]VERSION ones
>     break :blk b.addLibrary(.{
>         .linkage = .dynamic,
>         .name = "gc",
>         .root_module = gc_mod,
>     });
> } else blk: {
>     break :blk b.addLibrary(.{
>         .linkage = .static,
>         .name = "gc",
>         .root_module = gc_mod,
>     });
> };
> ```
> 
> I would now write this:
> ...
> 
> This also means we can inline `gc_mod` as it is only used once:
> 
> ```zig
> // TODO: convert VER_INFO values to [SO]VERSION ones
> const gc = b.addLibrary(.{
>     .linkage = if (build_shared_libs) .dynamic else .static,
>     .name = "gc",
>     .root_module = b.createModule(.{
>         .target = target,
>         .optimize = optimize,
>     }),
> });
> ```
> 
> This is a very common pattern: https://github.com/search?q=repo%3Aziglang%2Fzig%20%22.root_module%20%3D%20b.createModule(.%7B%22&type=code

It looks better, thank you, @linusg, I will apply it.

Issue #568, opened 2023-08-26
Support 16-bit targets


ivmai:
Although usefulness is unclear, technically it is possible to support targets with 16-bit data pointers (and probably having longer pointers for functions).

ivmai:
Here's unfinished work to port bdwgc to DOS (small and medium memory models). Base commit is 5839684457.

```
diff --git a/README.md b/README.md
index e1a66e53..513aeda8 100644
--- a/README.md
+++ b/README.md
@@ -285,8 +285,7 @@ Note that this may also be an issue with packed records/structs, if those
 enforce less alignment for pointers.
 
 A port to a machine that is not byte addressed, or does not use 32 bit
-or 64 bit addresses will require a major effort.  A port to plain MSDOS
-or win16 is hard.
+or 64 bit addresses will require a major effort.
 
 For machines not already mentioned, or for nonstandard compilers,
 some porting suggestions are provided [here](docs/porting.md).
diff --git a/backgraph.c b/backgraph.c
index e776577d..49a1e5a5 100644
--- a/backgraph.c
+++ b/backgraph.c
@@ -73,7 +73,11 @@ typedef struct back_edges_struct {
 
 /* Allocate a new back edge structure.  Should be more sophisticated    */
 /* if this were production code.                                        */
-#define MAX_BACK_EDGE_STRUCTS 100000
+#ifdef TINY_CONFIG
+# define MAX_BACK_EDGE_STRUCTS 200 /* some small value */
+#else
+# define MAX_BACK_EDGE_STRUCTS 100000
+#endif
 static back_edges *back_edge_space = 0;
 STATIC int GC_n_back_edge_structs = 0;
                                 /* Serves as pointer to never used      */
diff --git a/checksums.c b/checksums.c
index eea72de6..a26f9a14 100644
--- a/checksums.c
+++ b/checksums.c
@@ -17,8 +17,13 @@
 
 /* This is debugging code intended to verify the results of dirty bit   */
 /* computations.  Works only in a single threaded environment.          */
+#ifdef TINY_CONFIG
+# define NSUMS 1000
+# define OFFSET 0x1000
+#else
 # define NSUMS 10000
 # define OFFSET 0x10000
+#endif
 
 typedef struct {
         GC_bool new_valid;
diff --git a/cord/cordxtra.c b/cord/cordxtra.c
index 9786b5aa..801bc384 100644
--- a/cord/cordxtra.c
+++ b/cord/cordxtra.c
[...]

Issue #393, opened 2021-11-25
CMake building for bare metal


niansa:
Hey, which cmake options do I need to pass to build for bare metal?

ivmai:
I've not heard someone tried to build for bare metal using CMake. Is there any issue if no specific option is passed?

ivmai:
Are you using Musl?

niansa:
Nope, just some super-minimal C and C++ standard library without any posix stuff. Just malloc and free for memory management as well as some mmap/munmap-like functions.

ivmai:
It would be interesting to know the results of your experiments. Patches are welcomed too!

ivmai:
> Just malloc and free for memory management as well as some mmap/munmap-like functions.

The minimum you need is malloc (i.e. a way to get memory from the underlying subsystem).

A porting guide might be helpful for you - [porting.md](https://github.com/ivmai/bdwgc/blob/master/doc/porting.md)

PR #721, opened 2025-03-26
Add option to disable output entirely

aykevl:
This is mostly useful in environments where printf adds a non-negligible amount of code (for example, WebAssembly in the browser). Disabling all output shrinks the binary. Of course, this option should not be used when testing/debugging the GC.

I didn't test the binary size difference on WebAssembly, but on Linux (arm64) with a statically linked musl libc the difference is 20kB in my setup.

### DIFF
--- a/docs/README.macros
+++ b/docs/README.macros
@@ -578,6 +578,9 @@ GC_ANDROID_LOG (Android only)   Output error/debug information to Android log.
 CONSOLE_LOG (Win32 only)        Output error/debug information to stdout and
   stderr.
 
+GC_DISABLE_OUTPUT               Disable error/debug logging entirely. Useful to
+  reduce binary size but will also prevent any errors from being printed.
+
 GC_DONT_EXPAND  Do not expand the heap unless explicitly requested or forced
   to.
 
--- a/misc.c
+++ b/misc.c
@@ -2025,6 +2025,9 @@ GC_write(int fd, const char *buf, size_t len)
 /* Floating point arguments and formats should be avoided, since FP       */
 /* conversion is more likely to allocate memory.                          */
 /* Assumes that no more than BUFSZ-1 characters are written at once.      */
+#if defined(GC_DISABLE_OUTPUT)
+#define GC_PRINTF_FILLBUF(buf, format)
+#else
 #define GC_PRINTF_FILLBUF(buf, format)                      \
   do {                                                      \
     va_list args;                                           \
@@ -2035,10 +2038,12 @@ GC_write(int fd, const char *buf, size_t len)
     if ((buf)[sizeof(buf) - 1] != 0x15)                     \
       ABORT("GC_printf clobbered stack");                   \
   } while (0)
+#endif
 
 void
 GC_printf(const char *format, ...)
 {
+#if !defined(GC_DISABLE_OUTPUT)
   if (!GC_quiet) {
     char buf[BUFSZ + 1];
 
@@ -2056,6 +2061,7 @@ GC_printf(const char *format, ...)
     }
 #endif
   }
+#endif
 }
 
 void
@@ -2070,6 +2076,7 @@ GC_err_printf(const char *format, ...)
 void
 GC_log_printf(const char *format, ...)
 {
+#if !defined(GC_DISABLE_OUTPUT)
   char buf[BUFSZ + 1];
 
   GC_PRINTF_FILLBUF(buf, format);
@@ -2084,6 +2091,7 @@ GC_log_printf(const char *format, ...)
     ABORT("write to GC log failed");
   }
 #endif
+#endif // !defined(GC_DISABLE_OUTPUT)
 }
 
 #ifndef GC_ANDROID_LOG
@@ -2126,7 +2134,9 @@ void
 GC_err_puts(const char *s)
 {
   /* Note: write errors are ignored.  */
+  #if !defined(GC_DISABLE_OUTPUT)
   (void)WRITE(GC_stderr, s, strlen(s));
+  #endif
 }
 
 STATIC void GC_CALLBACK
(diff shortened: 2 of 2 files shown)


ivmai:
Okay!

ivmai:
BTW. Could you please reference your project where the feature is needed?

aykevl:
I'm using this library in TinyGo. Right now we only use bdwgc on Linux, MacOS, and Windows (where binary size doesn't matter a lot), but I'm working on WebAssembly support. For WebAssembly I'd like to keep binary size down as much as possible.

See: https://github.com/tinygo-org/tinygo/pull/4812

aykevl:
_Maybe_ this library could even be used baremetal? That would be interesting to investigate. Our own GC is kinda slow. For baremetal (microcontrollers mostly), binary size matters even more.

ivmai:
> _Maybe_ this library could even be used baremetal? That would be interesting to investigate. Our own GC is kinda slow. For baremetal (microcontrollers mostly), binary size matters even more.

Yes, it should be possible, but I've not heard of such experiments. See NOSYS macro.

ivmai:
> I'm using this library in TinyGo. Right now we only use bdwgc on Linux, MacOS, and Windows (where binary size doesn't matter a lot), but I'm working on WebAssembly support. For WebAssembly I'd like to keep binary size down as much as possible.

Off-topic: BTW, ages ago my record was a 55KB Win23 executable containing a GC java benchmark (GCBench) compiled to native code, java runtime and bdwgc.

ivmai:
Note: I've tried to produce a minimal statically-linked executable w/o proposed patch:
Target / compiler : Cygwin64 / clang-8
Command: `clang -I include -Os -Wall -Wextra -Wpedantic -D NO_DEBUGGING -D SMALL_CONFIG -D GC_NO_FINALIZATION -D NO_CLOCK
-D NO_GETENV -D IGNORE_DYNAMIC_LOADING -ffunction-sections -fdata-sections -s -o hugetest.exe tests/huge.c extra/gc.c -Wl,--gc-sections -Wl,-Map,hugetest.map`
Executable file size: 49KB

ivmai:
Somewhat related issue: #568

aykevl:
> > _Maybe_ this library could even be used baremetal? That would be interesting to investigate. Our own GC is kinda slow. For baremetal (microcontrollers mostly), binary size matters even more.
> 
> Yes, it should be possible, but I've not heard of such experiments. See NOSYS macro.
> Related open issue: #393

Ah, cool, thank you! Will look into NOSYS when I get around to it.

aykevl:
> Off-topic: BTW, ages ago my record was a [55KB Win23 executable](https://www.ivmaisoft.com/jcgo/demos.htm) containing a GC java benchmark (GCBench) compiled to native code, java runtime and bdwgc.

Right now in TinyGo (with this patch included), I've got a binary of 39kB on linux/arm64:

```
$ tinygo build -o test ./testdata/gc.go
$ strip test
$ ls -l test
.rwxr-xr-x@ 39k ayke 29 mrt 09:24 test
```

This is a small and (overly) simple test to see whether the GC is working at all and not overwriting memory.

aykevl:
To be clear, the goal  of this patch is to avoid linking in the vsnprintf function which is rather large. TinyGo binaries do not normally call printf-like functions, Go has its own formatting code. The code size savings of bdwgc itself are small in comparson.

Before:

```
   code  rodata    data     bss |   flash     ram | package
------------------------------- | --------------- | -------
      0       0       0       7 |       0       7 | (padding)
    844    6968     340    4132 |    8152    4472 | (unknown)
  21716    2135     876  118147 |   24727  119023 | C bdwgc
   2900       0       0       0 |    2900       0 | C compiler-rt
  12992    2729       4     568 |   15725     572 | C musl
      0      32       0       0 |      32       0 | Go types
    436      24       0      16 |     460      16 | internal/task
    252      41       4     112 |     297     116 | main
   1532     217       0      18 |    1749      18 | runtime
------------------------------- | --------------- | -------
  40672   12146    1224  123000 |   54042  124224 | total
```

After:

```
   code  rodata    data     bss |   flash     ram | package
------------------------------- | --------------- | -------
      0       0       0       7 |       0       7 | (padding)
    628    5262     344    4308 |    6234    4652 | (unknown)
  21288    2039     872  118147 |   24199  119019 | C bdwgc
   4476      27       4     568 |    4507     572 | C musl
      0      32       0       0 |      32       0 | Go types
    436      24       0      16 |     460      16 | internal/task
    252      41       4     112 |     297     116 | main
   1532     217       0      18 |    1749      18 | runtime
------------------------------- | --------------- | -------
  28612    7642    1224  123176 |   37478  124400 | total
```


(The output is a bit weird and designed for MCU usage, "flash" basically means binary size and "ram" means static data like non-constant globals).

The bdwgc library goes from 24727 to 24199 bytes, a difference of half a kilobyte. That's not very large. However, the musl libc (which is statically linked) goes from 15725 to 4507 - around 10kB reduction. And not everything is accounted for: `(unknown)` also reduces a bit (the binary size attribution code isn't perfect, it bases itself on DWARF data and other things after linking so can't always correctly attribute where some binary data originally comes from).

WebAssembly will be different, but since we also use a statically linked binary on WebAssembly, I expect a similar binary size reduction.

(I also see that compiler-rt is ~3kB with vsnprintf linked in, that's all due to 128-bit float support and won't apply to WebAssembly).

ivmai:
Got it, now it is more clear.
I will process the patch this week.

deadprogram:
I am eagerly awaiting this PR for https://github.com/tinygo-org/tinygo/pull/4812 :smile_cat:

deadprogram:
@aykevl please note some CI failures in this PR such as https://ci.appveyor.com/project/ivmai/bdwgc/builds/51857117/job/n589bbg0ahctoyg6

ivmai:
> I am eagerly awaiting this PR for [tinygo-org/tinygo#4812](https://github.com/tinygo-org/tinygo/pull/4812) 😸

Yes, it is in my nearest plans.

ivmai:
Finally done. I have changed the concept - it now prints the format string as it instead of not printing at all. (This is done for the simplicity of the change as otherwise, we will need to remove all unused functions/vars to avoid "defined but not used" warnings.)

Should work for you - please test.

PR #730, opened 2025-04-21
Document functions/macros' out-of-memory behavior.

### LINKED ISSUE #729: Documentation request: pointers returned in out-of-memory situations
The comment block just before the declaration of GC_oom_func in include/gc/gc.h says:

   "When there is insufficient memory to satisfy an allocation request,  we return (*GC_oom_fn)(size)."

However, the various macros and functions that users of the library invoke to allocate memory do not say whether this guarantee applies to them, or not.

The reason why this is a (small) issue for me is the following.

- I can see that the code of the library, e.g. in GC_generic_malloc, tests whether the allocation failed, and if so, it returns the result of an invocation of GC_oom_func.

- Users can provide their own code for GC_oom_func.

- In my use case, since there is no useful recovery possible in out-of-memory situations, the function I supply will abort the program.

- Given this fact, there would seem to be no need to follow calls to e.g. GC_malloc with code to test the return value and abort the program if it is NULL, *provided* that GC_malloc guarantees that it won't return NULL unless GC_oom_func does so.

I would like to optimize away this post-call tes. Therefore I would like to request that for every user-visible macro and function that does allocation, you add a sentence to its description that says something like:

   "This macro/function is guaranteed not to return NULL, unless the current value of GC_oom_func, when invoked on the allocation size, itself returns NULL."

or possibly

   "This macro/function is guaranteed to always return either the non-NULL address of a properly allocated memory block, or, if that is not possible, the result of invoking GC_oom_func on the allocation size. The default GC_oom_func returns NULL, but users can change this default by calling GC_set_oom_fn()."

I *could* check the entire call tree of allocation macros and function to see whether on all possible execution paths, they guarantee to call GC_oom_func (actually, *GC_get_oom_fn()) instead of returning a NULL pointer, but that would only tell me whether this is the case in the *current* version of the library. I would prefer to optimize away the post-call check only if the project owners intend to maintain such guarantees in future versions as well.

Thanks in advance.

zsomogyi:
Document the behavior of alloocation functions and macros in out-of-memory cases.

Closes #729.

### DIFF
--- a/include/gc/gc.h
+++ b/include/gc/gc.h
@@ -125,6 +125,11 @@ GC_API void GC_CALL GC_set_markers_count(unsigned);
 /* By default, this just returns NULL.  GC_oom_fn must not be 0.  Both  */
 /* the setter and the getter acquire the allocator lock (in the reader  */
 /* mode in case of the getter) to avoid data race.                      */
+/* If you set GC_oom_fn to point to a function that never returns NULL, */
+/* probably by aborting the program instead, then invocations of the    */
+/* functions and macros below marked as "never returning NULL unless    */
+/* GC_oom_fn returns NULL" won't need to be followed by code that       */
+/* checks for NULL.                                                     */
 typedef void *(GC_CALLBACK *GC_oom_func)(size_t /* bytes_requested */);
 GC_API GC_ATTR_DEPRECATED GC_oom_func GC_oom_fn;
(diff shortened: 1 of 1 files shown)


ivmai:
I've done a number of changes to the PR. Merged now, but please review, just in case I've missed something important for you.

zsomogyi:
I reviewed the committed diff, and your changes are perfectly fine. Thank you.

Issue #726, opened 2025-04-13
disclaim_bench failed because of Too many retries in GC_alloc_large


ivmai:
Source: master (6091f553e)
Build: https://github.com/ivmai/bdwgc/actions/runs/14428080842/job/40459590233
Host: MacOS
Command: `zig build -Dtarget=native -DBUILD_SHARED_LIBS=true -Dbuild_cord=true -Ddisable_handle_fork=false -Denable_cplusplus=true -Denable_gc_assertions=true -Denable_gc_debug=false -Denable_large_config=false -Denable_munmap=true -Denable_parallel_mark=true -Denable_redirect_malloc=false -Denable_rwlock=true -Denable_thread_local_alloc=true -Denable_threads=true -Denable_werror test`
Output (disclaim_bench): Too many retries in GC_alloc_large

ivmai:
Related issue #691. See https://github.com/ivmai/bdwgc/issues/691#issuecomment-2663700518 - a check was added in that commit with the limit (MAX_ALLOCLARGE_RETRIES) of 3 retries. Is it the cases described in that issue or just small limit?

tholp:
We receive "Too many retries in GC_alloc_large" in certain 32 bit programs.  In our cases it seems that the problem disappears when `MAX_ALLOCLARGE_RETRIES` is set to a number larger than `10`.

I do not know what these "retries" are, but `10` seems like some kind of arbitrary value.

ivmai:
Okay, could you please make a PR? But I will keep this issue open because "10 seems like some random value"

tholp:
OK.
(Actually we have set the value to 100 because we didn't have any problems before the test was introduced).  But I will set it to 10 in the pull request.

ivmai:
Source: master (ed9dac80b)
Build: https://app.travis-ci.com/github/ivmai/bdwgc/jobs/634261349
Config: `CFLAGS_EXTRA="-O3 -D _FORTIFY_SOURCE=2 -D DONT_PROTECT_PTRFREE -D FORCE_MPROTECT_BEFORE_MADVISE -D GC_UNMAPPED_REGIONS_SOFT_LIMIT=120" CONF_OPTIONS="--enable-gc-assertions --enable-cplusplus --enable-static" GCTEST_WITH_MPROTECT_VDB=true`
Output (gctest):
```
 GC_USE_GETWRITEWATCH=0 ./gctest; fi
Running on LINUX/POWERPC target
[...]

PR #740, opened 2025-07-22
GC_alloc_large: increase MAX_ALLOCLARGE_RETRIES (to 10)

tholp:
We experience _Too many retries in GC_alloc_large_ in some 32 bit programs.  Increasing `MAX_ALLOCLARGE_RETRIES` solves the problem. 

See [disclaim_bench failed because of Too many retries in GC_alloc_large](https://github.com/ivmai/bdwgc/issues/726#issuecomment-3097607685).

### DIFF
--- a/malloc.c
+++ b/malloc.c
@@ -46,7 +46,7 @@ GC_alloc_reclaim_list(struct obj_kind *ok)
 STATIC ptr_t
 GC_alloc_large(size_t lb_adjusted, int kind, unsigned flags, size_t align_m1)
 {
-#define MAX_ALLOCLARGE_RETRIES 3
+#define MAX_ALLOCLARGE_RETRIES 10
   int retry_cnt;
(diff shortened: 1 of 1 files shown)


tholp:
A lot of checks fails, but I don't see that it can be due to this change.

ivmai:
> A lot of checks fails, but I don't see that it can be due to this change.

Right, GH actions are not working, because of "account payments have failed or your spending limit needs to be increased".
(I'm using Free plan, but it seems there's an issue with the budget - I haven't figure how to solve it yet)

ivmai:
Relates to issue #726 (but does not solve it completely)

ivmai:
Merged as ed9dac80b

ivmai:
PS. As mentioned in #726, a 32-bit test case failed once even after increasing the limit to 10.

PR #750, opened 2025-08-16
build.zig: Switch to unmanaged containers

linusg:
This is in preparation for the upcoming Zig 0.15 release. Managed containers (storing an allocator) will go away, unmanaged ones (passing an allocator to functions that need to allocate) are becoming the default. What is called `std.ArrayListUnmanaged` up until Zig 0.14 has been renamed to `std.ArrayList` in 0.15 (https://github.com/ziglang/zig/pull/24801) with the old name still available as a deprecated alias.

This change is backwards compatible as both types of containers have existed for a long time.

PR #751, opened 2025-08-25
Switch to unmanaged containers in Zig script

ivmai:
Supersedes PR #750.

This is in preparation for the upcoming Zig 0.15 release. Managed containers (storing an allocator) will go away, unmanaged ones (passing an allocator to functions that need to allocate) are becoming the default.  What is called `std.ArrayListUnmanaged` up until Zig 0.14 has been renamed to `std.ArrayList` in Zig 0.15 with the old name still available as a deprecated alias.

This change is backwards compatible as both types of containers have existed for a long time.


ivmai:
Tested locally (on rv64) with Zig 0.14.0 and 0.15.1.

PR #753, opened 2025-09-09
Workaround some gcc-13 analyzer warnings

ivmai:
Workaround "use of unitialized variable" false positives (FPs) reported by gcc-13 `-fanalyzer`:
1. In .cc files
2. In tests/leak.c

Use Ubuntu noble for analyzer builds where no more FPs.


ivmai:
Builds are OK

PR #760, opened 2025-10-04
Use Zig build system's standard way to set linking mode

congcongkuo:
As we have been talking at [](https://github.com/bdwgc/bdwgc/issues/723)

### DIFF
--- a/.github/workflows/zig-build.yml
+++ b/.github/workflows/zig-build.yml
@@ -5,7 +5,7 @@ on: [ push, pull_request ]
 
 jobs:
   build:
-    name: ${{ matrix.os }} ${{ matrix.ttriple }} thr:${{ matrix.enable_threads }} rwlock:${{ matrix.enable_rwlock }} redir:${{ matrix.redirect_malloc }} gcdeb:${{ matrix.enable_gc_debug }} munmap:${{ matrix.enable_munmap }} paramark:${{ matrix.parallel_mark }} thrlocal:${{ matrix.thread_local_alloc }} dll:${{ matrix.shared_libs }}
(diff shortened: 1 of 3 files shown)


congcongkuo:
There is a TODO comment https://github.com/bdwgc/bdwgc/blob/1a3644efffc187e33734ece247ad3bb6b6b08522/build.zig#L386-L387
releted to the variable `build_shared_libs` that is replaced in this PR. This TODO comment maybe need to be updated and I can't realize what this comment are taling about.

ivmai:
> There is a TODO comment

> // TODO: declare that the libraries do not refer to external symbols of `build_shared_libs`. 
> related to the variable build_shared_libs that is replaced in this PR. This TODO comment maybe need to be updated and I can't realize what this comment are tailing about.

We try to keep same behavior across all build scripts, and if some functionality is implemented in `configure.ac`, then we implement it in `CMakeLists.txt` and in `build.zig`, if applicable, of course.
The mentioned TODO says that the feature existing in `CMakeLists.txt` (added in commit 813780111) is not yet implemented in `build.zig`. When we added Zig support, we used CMake file as a basis and just added this TODO instead of figuring out how to pass `--no-undefined` option to linker and whether such option was needed actually. If I remember correctly, `-Wl,--no-undefined` was needed to check that certain non-inlined atomic operations references `libatomic.so` (or `libatomic_ops.so` when gcc built-in atomic operations are not available). The case about option `--no-undefined` is explained, e.g., in https://stackoverflow.com/questions/2356168/force-gcc-to-notify-about-undefined-references-in-shared-libraries .

@congcongkuo, if you are sure that `--no-undefined` is useless with Zig (e.g. implicitly provided as we have with autotools), then let's remove this TODO.

ivmai:
Closing in favor of PR #761

PR #761, opened 2025-10-05
Use Zig build-system standard way to set linkage mode (replaces PR 760)

ivmai:
This replaces PR #760 by @congcongkuo


ivmai:
@congcongkuo, regarding the [TODO item](https://github.com/bdwgc/bdwgc/pull/760#issuecomment-3368550121), could you please check if option `-Wl,--no-undefined` is meaningful for Zig build system?

PR #762, opened 2025-10-05
GH Actions: Update zig to v0.15.1 final

ivmai:


ivmai:
Probably we could remove some FIXME/TODO items after update to 0.15.1. Let me check it.

ivmai:
> Probably we could remove some FIXME/TODO items after update to 0.15.1. Let me check it.

Only issues for `loongarch64-linux-gnusf` and `powerpc64-linux-gnu` have been resolved since previous Zig major version.

PR #763, opened 2025-10-10
Allow the GC state to be re-initialized (in some configurations)

ivmai:
This is upstreaming of https://github.com/bdwgc/bdwgc/commit/8b95edb
Plus my supplementary commits like transfer of global/static vars to GC_arrays.


ivmai:
Test are OK (no regressions observed).

PR #793, opened 2025-10-31
Fix reliance on glibc-specific `__data_start` for some Linux platforms

Expertcoderz:
Setting `DATASTART` based on `__data_start` only works when the `__data_start` symbol is defined somewhere, and this seems to be a glibc-specific feature (i.e. not available on musl libc). To avoid a "Wrong DATASTART/END pair" runtime error[^1] when bdwgc is built for non-glibc Linux platforms, we should fall back to `SEARCH_FOR_DATA_START` if not using glibc.

Affected architectures:
- aarch64
- arc
- loongarch
- mips
- nios2
- or1k
- riscv
- s390
- tilegx
- tilepro

Compiled code remains unaffected when using glibc.

[^1]: Side note: the linker doesn't seem to catch the missing definition at link time due to `__data_start` being declared as a weak symbol.

### DIFF
--- a/include/private/gcconfig.h
+++ b/include/private/gcconfig.h
@@ -1680,9 +1680,13 @@ extern char *_STACKTOP;
 #  define MACH_TYPE "LOONGARCH"
 #  define CPP_WORDSZ (__SIZEOF_SIZE_T__ * 8)
 #  ifdef LINUX
-#    pragma weak __data_start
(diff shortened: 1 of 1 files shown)


ivmai:
Okay!

ivmai:
Replaced by PR #796

Issue #288, opened 2019-06-29
gctest or threadleaktest hang sometimes on Ubuntu/x64 if --enable-gc-debug --enable-redirect-malloc


ivmai:
Build link: https://travis-ci.org/ivmai/bdwgc/jobs/551952193
Host: Linux/x64 (Xenial)
Source: https://github.com/ivmai/bdwgc/commit/d7e2c143c740a93cea9acb397c2e3452327eb0ec (release-7.6)
How to reproduce: configure --enable-redirect-malloc --enable-gc-debug --enable-cplusplus --enable-gc-assertions && make check

ivmai:
Latest failed build: https://travis-ci.org/ivmai/bdwgc/jobs/576996250

ivmai:
Unable to reproduce locally.

ivmai:
Not observed on other GC versions (including master).
Had not been observed before update to Travis Xenial image.

ivmai:
Not observed with --disable-static

ivmai:
Source: master (3f419b6)
Build: https://app.travis-ci.com/github/ivmai/bdwgc/jobs/604837855
Host: Ubuntu 22.04.2 LTS / x64
Compiler: gcc 11.3.0
How to build: ./configure --enable-redirect-malloc --enable-gc-debug --enable-cplusplus --enable-gc-assertions && make -j check CFLAGS_EXTRA="-D _FORTIFY_SOURCE=2"
Occurrence: very rare

ivmai:
Source: master (b678225)
Build: https://app.travis-ci.com/github/ivmai/bdwgc/jobs/607405872
Host: Ubuntu 22.04.2 LTS / x64

ivmai:
Source: master (5839684)
Build: https://app.travis-ci.com/github/ivmai/bdwgc/jobs/608674865
Host: Linux/amd64
Config: CFLAGS_EXTRA="-D _FORTIFY_SOURCE=2" CONF_OPTIONS="--enable-redirect-malloc --enable-gc-debug --enable-cplusplus --enable-gc-assertions"

ivmai:
Hang in threadleaktest.
Build: https://app.travis-ci.com/github/ivmai/bdwgc/jobs/609180209
Source: master (6f8349a)
Host: Ubuntu 22.04.2 LTS / amd64 (Jammy)
Compiler: gcc (Ubuntu 11.3.0-1ubuntu1~22.04.1) 11.3.0
[...]

PR #794, opened 2025-10-31
Fix hang if malloc redirection combined with gc-debug

ivmai:
Fix should fix issues #288, at least.

Also, contains some refactoring.


ivmai:
Builds seem to be good (no regression)

PR #798, opened 2025-11-03
Fix code defects in GC_finalize_all and enable_incremental_mode

ivmai:


ivmai:
Builds seems to be OK

PR #805, opened 2025-11-14
Fix integration with Coveralls by using new reporter tool

ivmai:
Same as https://github.com/bdwgc/libatomic_ops/pull/94


ivmai:
Builds are OK. Coveralls reporting is OK

PR #813, opened 2025-11-29
Fix comment for endif in GC_print_callers

ivmai:
Remove wrong (outdated) comment for `endif` belonging to `ifdef CALLINFO_USE_ADDR2LINE`.


ivmai:
No builds needed.

PR #826, opened 2025-12-21
Specify SO-version for shared libraries built by Zig

congcongkuo:
Issue https://github.com/bdwgc/bdwgc/issues/638 (bdwgc).

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -16,7 +16,10 @@ const std = @import("std");
 // The Zig version here should match that in file `build.zig.zon`.
 const zig_min_required_version = "0.14.0";
 
-// TODO: specify `PACKAGE_VERSION` and `LIB*_VER_INFO`.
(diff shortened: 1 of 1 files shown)


congcongkuo:
Or maybe we can write such as
```
const LIBGC_VER_INFO = "6.3.5";
.version = std.SemanticVersion.parse(LIBGC_VER_INFO) catch unreachable,
```

ivmai:
> Or maybe we can write such as
> 
> ```
> const LIBGC_VER_INFO = "6.3.5";
> .version = std.SemanticVersion.parse(LIBGC_VER_INFO) catch unreachable,
> ```

I have already solved the problem, using libtool version, however I like your idea to use the sem ver notation.
But we need same notation in Makefile.am and cord.am.
If you could update your patch implementing it properly for .am files, I  am ready to use you patch.
Thank for the contribution anyway!

PR #830, opened 2025-12-23
Don't close thread handles on dllmain process exit

tobil4sk:
reserved != NULL indicates that the process is exiting, in which case handles will be closed automatically so we shouldn't do that here.

See: https://devblogs.microsoft.com/oldnewthing/20120105-00/?p=8683

> Don’t worry about freeing memory; it will all go away when the process
> address space is destroyed. Don’t worry about closing handles; handles
> are closed automatically when the process handle table is destroyed.

### DIFF
--- a/win32_threads.c
+++ b/win32_threads.c
@@ -1991,7 +1991,6 @@ GC_DllMain(HINSTANCE inst, ULONG reason, LPVOID reserved)
   thread_id_t self_id;
 
   UNUSED_ARG(inst);
-  UNUSED_ARG(reserved);
   /*
    * Note that `GC_use_threads_discovery` should be called by the client
    * application at start-up to activate automatic thread registration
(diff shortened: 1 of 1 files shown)


ivmai:
Okay, I've reviewed the code, refactored it a bit and pushed to https://github.com/ivmai/bdwgc-fork/tree/refs/heads/pr/fix-dllmain-830 for testing.
Have you seen the crashes w/o this patch?

tobil4sk:
> Have you seen the crashes w/o this patch?

I haven't experienced any myself. The reason for this patch is because I was concerned about how to properly handle `DLL_PROCESS_DETACH` for #831, and this seems to be the correct way to handle things in general according to microsoft docs, so I thought I'd split this out into a standalone patch.

ivmai:
> See: https://devblogs.microsoft.com/oldnewthing/20120105-00/?p=8683

Here's same info from MS Learn resource (I'm mentioning this to have yet another explanation):
https://learn.microsoft.com/en-us/windows/win32/dlls/dllmain

ivmai:
I will merge it in another PR. Thank you!

PR #906, opened 2026-04-11
Fix dll loader deaklock

ivmai:
* New macro TEST_NO_THREADS_DISCOVERY to test without use_threads_discovery
* Fix deadlock caused by MessageBox call while DllMain thread is suspended (this closes #886)
* Minimize code duplication in ABORT definition (refactoring) 
* Fix 'GC_DllMain undeclared' compiler error if CPPCHECK defined


ivmai:
No regression

PR #923, opened 2026-05-02
Deadlock in mutithread mode under Windows when using Zig toolchain

ebezault:
I'm the main developer of an open source compiler for the Eiffel programming language (https://github.com/gobo-eiffel/gobo). It translates the Eiffel code to C, which is then compiled using the Zig toolchain (which itself uses the clang compiler). The compiled Eiffel programs also use the Boehm-Demers-Weiser GC.

Under Windows, when compiled with Zig toolchain, `signal` indirectly calls `RtlAllocateHeap` which uses mutual exclusion when accessing the heap. If some other thread was in the middle of calling `RtlAllocateHeap` when it was suspended (stop the world), this ends up in a deadlock.

Using `SetUnhandledExceptionFilter` instead of `signal` solves the deadlock issue, but this function is extremely slow when compiled with Zig toolchain. 

I'm not sure what the best solution is to solve this issue, but my current workaround (to avoid the deadlock, without slowing down the execution), is to avoid setting the fault handler in that case when threads have already been created.

### DIFF
--- a/os_dep.c
+++ b/os_dep.c
@@ -1038,9 +1038,27 @@ GC_set_and_save_fault_handler(GC_fault_handler_t h)
 #      endif
 #    endif /* !IRIX5 || !THREADS */
 #  else
-  old_segv_hand = signal(SIGSEGV, h);
+  /*
+   * Under Windows when compiling with zig/clang, 'signal'
+   * indirectly calls 'RtlAllocateHeap' which uses mutual
+   * exclusion when accessing the heap. If some other thread
+   * was in the middle of calling 'RtlAllocateHeap' when it
+   * was suspended (stop the world), this ends up in a
+   * deadlock. Using 'SetUnhandledExceptionFilter' instead of
+   * 'signal' solves the deadlock issue, but this function is
+   * extremely slow when compiled with zig/clang. So do not
+   * set fault handler in that case when threads have already
+   * been created.
+   */
+#    if defined(GC_WIN32_THREADS) && defined(__clang__)
+  if (!GC_need_to_lock) {
+#    endif
+    old_segv_hand = signal(SIGSEGV, h);
 #    ifdef HAVE_SIGBUS
-  old_bus_hand = signal(SIGBUS, h);
+    old_bus_hand = signal(SIGBUS, h);
+#    endif
+#    if defined(GC_WIN32_THREADS) && defined(__clang__)
+  }
 #    endif
 #  endif /* !USE_SEGV_SIGACT */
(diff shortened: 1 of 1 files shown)


ivmai:
> my current workaround (to avoid the deadlock, without slowing down the execution), is to avoid setting the fault handler in that case when threads have already been created.

It seems this is equivalent to undefine `WRAP_MARK_SOME`, right?

ivmai:
As commit 29be14a4a says about the need of `WRAP_MARK_SOME`:
```
  /* Windows 98 appears to asynchronously create and remove writable   */
  /* memory mappings, for reasons we haven't yet understood.  Since    */
  /* we look for writable regions to determine the root set, we may    */
  /* try to mark from an address range that disappeared since we       */
  /* started the collection.  Thus we have to recover from faults here. */
  /* This code does not appear to be necessary for Windows 95/NT/2000. */
  /* Note that this code should never generate an incremental GC write */
  /* fault.                                                            */
```
It looks like we turn off `WRAP_MARK_SOME` for `MSWIN32`

ebezault:
Thank you for your feedback. To be complete, I have no problem when using the msvc C compiler on Windows (I'm on Windows 11). My problem is when using the Zig toolchain for C compilation, which is using the clang C compiler and (I think) the C libraries from mingw. In these C libraries, `signal` indirectly calls `RtlAllocateHeap` which uses mutual exclusion when accessing the heap. If some other thread was in the middle of calling `RtlAllocateHeap` when it was suspended (stop the world), this ends up in a deadlock. 
Screenshot 2026-05-04 231119
Screenshot 2026-05-04 230851

That being said, I tried to force the definition of `WRAP_MARK_SOME` and I still have the deadlock. I then made sure that `WRAP_MARK_SOME` is not defined and so far I was not able to get the deadlock. So I think that we can get rid of the workaround I provided in this Pull Request. Instead of that I will make sure that:

- `WRAP_MARK_SOME` is defined when using msvc on Windows. It looks like it is the default, so nothing needs to be done.
- `WRAP_MARK_SOME` is not defined when using Zig toolchain (i.e. clang) or mingw (or any other C compiler using the mingw C libraries) on Windows. For that I will use `NO_WRAP_MARK_SOME` as per this [line](https://github.com/bdwgc/bdwgc/blob/ea69934ed214cfb4e38d2a0176cc392af8055a83/include/private/gcconfig.h#L3315) in `gcconfig.h`.

ivmai:
Closing this PR in favor of alternate solution, further discussion in issue #924

PR #925, opened 2026-05-07
Fix deadlock in signal on Windows by using VEH

### LINKED ISSUE #924: Deadlock in RtlAllocateHeap on Windows when using Zig
As reported by @ebezault in #923:
Under Windows, when compiled with Zig toolchain, signal indirectly calls RtlAllocateHeap which uses mutual exclusion when accessing the heap. If some other thread was in the middle of calling RtlAllocateHeap when it was suspended (stop the world), this ends up in a deadlock.
Using SetUnhandledExceptionFilter instead of signal solves the deadlock issue, but this function is extremely slow when compiled with Zig toolchain.

To be complete, I have no problem when using the msvc C compiler on Windows (I'm on Windows 11). My problem is when using the Zig toolchain for C compilation, which is using the clang C compiler and (I think) the C libraries from mingw. In these C libraries, signal indirectly calls RtlAllocateHeap which uses mutual exclusion when accessing the heap. If some other thread was in the middle of calling RtlAllocateHeap when it was suspended (stop the world), this ends up in a deadlock.

(I'm the main developer of an open source compiler for the Eiffel programming language (https://github.com/gobo-eiffel/gobo). It translates the Eiffel code to C, which is then compiled using the Zig toolchain (which itself uses the clang compiler). The compiled Eiffel programs also use the Boehm-Demers-Weiser GC.)

ivmai:
This should close #924

### DIFF
--- a/.github/workflows/cmake-build.yml
+++ b/.github/workflows/cmake-build.yml
@@ -14,7 +14,7 @@ jobs:
       fail-fast: false
 
       matrix:
-        os: [ macos-15-intel, macos-latest, ubuntu-24.04-arm, ubuntu-latest, windows-11-arm, windows-latest ]
+        os: [ macos-26-intel, macos-latest, ubuntu-24.04-arm, ubuntu-latest, windows-11-arm, windows-latest ]
         c_compiler: [ cl, clang, gcc ]
         cxx_compiler: [ cl, clang++, g++ ]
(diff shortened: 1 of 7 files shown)


ivmai:
More work (and checking) is needed in PR, e.g. cooperation with GC_set_write_fault_handler.
Closing as unmerged for now.

PR #930, opened 2026-05-23
Expose nanosecond fraction of GC total time measurement

kylc:
Currently GC collection time measurements are only exposed with millisecond precision. This adds an API call to retrieve the nanosecond part.

In my usecase GC collection times are often on the order of a few milliseconds, so being limited to millisecond precision is too coarse.

### DIFF
--- a/alloc.c
+++ b/alloc.c
@@ -73,6 +73,12 @@ GC_get_full_gc_total_time(void)
   return GC_full_gc_total_time;
 }
 
+GC_API unsigned int GC_CALL
+GC_get_full_gc_total_time_ns_frac(void)
+{
+  return GC_full_gc_total_ns_frac;
+}
+
 GC_API unsigned long GC_CALL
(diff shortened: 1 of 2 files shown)


ivmai:
I have reworked the patch a bit, see commit a3a68ad:
- renamed the function to GC_get_full_gc_total_ns_frac
- changed return type to unsigned long (it's a design consideration to be in sync with GC_timeval_s)
- refined comment, added a call from gctest.

Please note that you should call GC_get_full_gc_total_time and GC_get_full_gc_total_ns_frac inside a single callback of GC_call_with_reader_lock, otherwise you will potentially get an non-atomic value of time if your app is multi-threaded.

ivmai:
Moved to PR #931

jeaye:
Thanks for the rework, merge, and tips, Ivan! We'll need to make sure we're using that lock, @kylc.

PR #931, opened 2026-05-25
Provide API to get nanosecond fraction of total GC time

ivmai:


ivmai:
No regression

kylc:
Thank you!

PR #957, opened 2026-09-15
Remove redundant FREE_PROFILER_HOOK calls in case of small objects plus some related refactoring

ivmai:


ivmai:
Commit https://github.com/bdwgc/bdwgc/pull/957/commits/48756b557839d35080b2374f42892a8473d95c8d relates to RP #621

ivmai:
Hello @jacob-hughes,
I have reworked your code a bit (see commit 48756b557839) to avoid redundant hook calls when freeing small objects - no it should be just one hook call for an object. Please test it.

PR #961, opened 2026-09-18
Handle userfaultfd protection ranges crossing VMAs

plajjan:
Linux 6.1 rejects UFFDIO_WRITEPROTECT ranges that cross VMAs, even when the entire range is registered. Retry these requests in smaller page-aligned pieces, preserving the existing error when a single page fails.

A GC mapping split with MADV_DONTDUMP reproduces the failure and passes with this change; a deliberately unregistered page still fails. The previously failing gctest and disclaimtest also pass with assertions enabled.


ivmai:
Okay, the PRs look good at first glance, all tests pass, I will review them in detail an merge them in a PR. Thank you!

